PR #2, opened 2017-11-06
Set Hellbender DrawScale

GreatEmerald:
The hellbender seems too small, set the size to 1.0 like the other models.


GreatEmerald:
This works, but might be worth checking the size of all vehices at some point, as in UT3 the DrawScale differs between models.

100GPing100:
We've got some activity here! Nice to see this!

GreatEmerald:
Yea, some people are interested in getting a few of the remaining bugs fixed, and this also makes the mod compile correctly.
I'll put up some of the issues that have been reported to me on the bug tracker too.

PR #20, opened 2017-12-11
Update UT3Hoverboard.uc

HellDragon-HK:
Fix team lock problem blue has, I tested this in-game and it works but I think it removes team lock from the hoverboard completely but as it's a weapon spawn and not a vehicle spawn on the field not much of an issue maybe?


GreatEmerald:
Thanks.
I'm not too sure about whether it's an issue or not. Technically it should never appear on its own. One way to ensure that would be to also make it self-destruct if it finds itself driverless for like a second.
@100GPing100 @Unre-Alex Any thoughts on that?

Unre-Alex:
It's not meant to be a standalone vehicle so it's not an issue to me. I've been using hoverboards with that unlock code ever since beta 4 release and have never seen one lying around in-game.

GreatEmerald:
Hm, all right then. It's better than the status quo anyway.

PR #21, opened 2017-12-11
Update UT3Goliath.uc

HellDragon-HK:
Fix Goliath to have 900 health like it does in UT3 from the 800 it currently has


GreatEmerald:
Confirmed, thanks.

PR #22, opened 2017-12-11
Update UT3Scorpion.uc

HellDragon-HK:
This restores the Scorpion's hatch animations for when it's sitting empty & in-use and also on eject automatically surprisingly.  Alex and I have both confirmed this works.

### DIFF
--- a/Classes/UT3Scorpion.uc
+++ b/Classes/UT3Scorpion.uc
@@ -455,6 +455,36 @@ simulated event KApplyForce(out Vector Force, out Vector Torque)
     }
 }
 
+function DriverLeft()
+{
+    GoToState('');
+    PlayAnim('hatch_open', 1.0, 0.1);
+
+    Super.DriverLeft();
+}
+
+event PostBeginPlay()
+{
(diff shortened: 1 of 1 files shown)


GreatEmerald:
Looks nice, but please also test online (local online match). I suspect that the `DriverLeft` function needs to be `simulated` or so to work online.

HellDragon-HK:
I don't play online at all, but would a lan game work although you said for the host it's the same as a bot match so...?

GreatEmerald:
Start a dedicated match and then join it with another instance of UT2004 on the same computer.

HellDragon-HK:
Alright give me a second to try it

HellDragon-HK:
Can't start a dedicated match, my PC is blocking it or something it opens the log window and then just vanishes soon after, this is happening across all installations modded and unmodded

HellDragon-HK:
Ok finally got it, no idea what was wrong but it does appear to work including self destruct, when getting out the timing was off once but it still played the animation.

GreatEmerald:
All right then.

PR #23, opened 2017-12-12
Driver Position Improvement for Manta

HellDragon-HK:
This is to raise the driver up so they don't clip the seat as much and move them closer to the controls, my own tests with it were with Kai, Asp, Satin, & Reaven character models which looked good to me and Alex also said he thought it was fine, I have no idea what model(s) he tested with.

Also what's this?
 We've detected the file encoding as ISO-8859-1. When you commit changes we will transcode it to UTF-8.

### DIFF
--- a/Classes/UT3Manta.uc
+++ b/Classes/UT3Manta.uc
@@ -1,7 +1,7 @@
 /*
- * Copyright © 2008 Wormbo
- * Copyright © 2012 100GPing100
- * Copyright © 2008, 2014 GreatEmerald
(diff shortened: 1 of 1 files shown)


Unre-Alex:
With UT3 models and UT2004 Mercenaries.

GreatEmerald:
OK, looks fine.

Don't worry about the encoding. It should be UTF-8, so GitHub just notifies that it will fix it in files that are not set to that for some reason.

PR #24, opened 2017-12-12
Improvements on Manta Plasma to be closer to UT3

HellDragon-HK:
Alex and I believe these values are closer to UT3 in how they feel, the only odd thing is the fire rate sound doesn't seem any faster in UT2004

In UT2004 mod another Manta is killed in 6 shots, in UT3 it takes exactly 8 shots on PC and exactly 7 on PS3, changing the value to 28 seems to now take 8 shots to kill another Manta like in UT3.


HellDragon-HK:
After further testing I think the UT3 speed values are the closer than what Alex and I decided on so it would be
Speed=2000
MaxSpeed=7000 which is the same in both games so could be removed I guess
AccelerationMagnitude 16000 which can also be removed probably

GreatEmerald:
Ooh, nice catch on the damage. This is due to:
http://greatemerald.xmpcommunity.com/uncodex/UT3-15/Source_utgamecontent/utdmgtype_mantabolt.html
See that 0.7 vehicle damage scaling? 36*0.7=25.2.

So the proper way to fix that is to create a new class `UTDmgType_MantaBolt` with vehicle and node damage scaling and apply it to the projectile.

As for the other settings, you concluded here that the numbers matching UT3 ones feel close as well, which is in line with using Occam's razor values for this. So stick to UT3 values precisely.

So edit this pull request with the needed changes and I'll squash them before merging.

HellDragon-HK:
I noticed the damage scaling yes, I'm checking all 3 codes of something at the same time when making changes.

So you want me to create a new class and ?  I'm lost sorry and it's above my ability I think but I can try I suppose.

GreatEmerald:
Yes. It's not very difficult at all. Check out https://github.com/GreatEmerald/UT3Vehicles/blob/master/Classes/UT3DmgType_LeviathanRocket.uc (you can use it as a template) and https://github.com/GreatEmerald/UT3Vehicles/blob/master/Classes/UT3LeviathanRocket.uc#L55 for how to apply it to a projectile.

HellDragon-HK:
I have no clue what I'm doing with code but I'll try

GreatEmerald:
That's how you learn! ;)
Make sure you test it in your installation, of course!

HellDragon-HK:
Well Alex and I both tried making a new code for respawn time and it didn't work, I tried (and commented about it) the SPMA and he tried the Cicada, neither time was affected

HellDragon-HK:
I did what you said and changed the damage to 36 with a scaling of 0.7 but now it's doing 54 damage

GreatEmerald:
Please include the `DamageType` changes in this pull request. You can edit more than just one file per pull request, after all.

HellDragon-HK:
Is that what you meant?

GreatEmerald:
No no, you removed the Manta projectile altogether! You should just upload a new file (go to your branch https://github.com/HellDragon-HK/UT3Vehicles/tree/patch-5, go to the Classes directory and press the Upload File button).

HellDragon-HK:
Ok thank you, I didn't know how to do that, last I tried it said I didn't have permission for it....uh can you automatically get it now or how does that work for those

GreatEmerald:
The DamageType looks OK but you deleted the entirety of the `Classes/UT3MantaPlasmaProjectile.uc` file. Readd that again.

This is all easier to do from git rather than GitHub, by the way...

PR #25, opened 2017-12-13
Update on Manta Ground Speed

HellDragon-HK:
1500 still feels a little faster than UT3 but it seems closer than 2000

### DIFF
--- a/Classes/UT3Manta.uc
+++ b/Classes/UT3Manta.uc
@@ -1,7 +1,7 @@
 /*
- * Copyright © 2008 Wormbo
- * Copyright © 2012 100GPing100
- * Copyright © 2008, 2014 GreatEmerald
+ * Copyright Â© 2008 Wormbo
+ * Copyright Â© 2012 100GPing100
+ * Copyright Â© 2008, 2014 GreatEmerald
  *
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions are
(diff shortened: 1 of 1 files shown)


GreatEmerald:
The units for speed are not necessarily equivalent between games. Model size in UU differs between the games, and thus so will velocity measurements. It would be better to test in e.g. Torlan how long it takes to get from one known point to another and then compare.

HellDragon-HK:
How would that work, Torlan in UT3 isn't the same size is it and it has extra paths :S

GreatEmerald:
Or does it?.. I looked at the code, and `EyeHeight` is the same between the games. Usually that's a good indicator of the model size difference between games. Base (`UT`/`X`)`Pawn` also have the same `GroundSpeed` and `AirSpeed`. So maybe 1 UU is in fact the same size between the games after all... As for how to test it, I'm not entirely sure either.

I suppose we could go with the assumption that measurement units match precisely, and look into that once we get some evidence to the contrary. That would make life much easier. In that case, I'll accept this.

It's funny that ground speed for Manta matters, considering that it's not really on the ground, but that's how the tank works too, so it's not that surprising.

HellDragon-HK:
I think what is going on is UT2004 has extra movement modifiers on vehicles such as I've noticed on wheeled vehicles with TransRatio which vastly effects speed but I haven't figured out how to fully work with it yet

PR #26, opened 2017-12-13
Manta Turn Speed change to be closer to UT3

HellDragon-HK:
It still feels like it might be a little slow but it's close, maybe change TurnTorqueMax to 185 or 190 still


GreatEmerald:
Hm, OK. I know it turns faster in UT3, but I'm not sure how fast. I'll just take it that your number is closer than the default.

PR #27, opened 2017-12-13
UT3 Manta Overall Handling Improvement

HellDragon-HK:
This is to help with the Manta's over tilting when it jumps, it looks closer to UT3 and doesn't feel like you're going to hit the ground sideways in my opinion


GreatEmerald:
Interesting. I make it roll more because in UT3 it rolls quite a bit, but I haven't considered damping.

However, please change the tab symbol to four spaces instead.

HellDragon-HK:
I'm sorry but I don't understand what you mean bu tab symbol

HellDragon-HK:
Now included in the Manta Full Game Play Improvement request so not needed

PR #28, opened 2017-12-13
UT3 Scorpion Fire Damage Degrade Closer to UT3

HellDragon-HK:
In UT3 vehicles degrade from fire damage much quicker than UT2004, this value seems to be a close match


GreatEmerald:
Hm, what does this value change exactly?

HellDragon-HK:
Fire destroys the vehicle quicker when it's on fire and empty which I didn't even know UT2004 had this feature until I looked

HellDragon-HK:
I don't understand how this is conflicting

HellDragon-HK:

PR #29, opened 2017-12-13
UT3 Scorpion Missing Collision and Explosion Sound Fix

HellDragon-HK:
This gives the UT3 Scorpion it's impact sounds (which at this time can only really be heard from weapon damage that throws the vehicle around as it doesn't seem to take impact damage from walls or vehicles), and gives it the correct explosion sound on death.  UAX file needs to be uploaded too but how?


GreatEmerald:
You can upload it to your fork's respective branch. At the top you see that it's `HellDragon-HK:patch-10`, which means https://github.com/HellDragon-HK/UT3Vehicles/tree/patch-10, here you can use the "upload file" from the menu if you don't want to clone it locally.

However, it is a binary file, so make sure that you will never need to edit it ever again!

HellDragon-HK:
Not sure on that as I've been wondering about attempting to increase the volume on some sounds, a new file can't be uploaded at a later date?

GreatEmerald:
The problem with binary files is that you can't store differences between their versions, which is what git does. So when you update a binary fire, you are not overwriting the one that already exists, but rather you are *adding* another file beside the one you replaced! The old file is simply hidden, but you still need to download it every time and it takes up disk space (and eventually GitHub stops allowing you to upload more because the size is exceeded).

One potential way to make this easier is to just not store any of the binary files on GitHub, but somewhere else, like Nextcloud or Google Drive, but that's an extra step needed for compiling the package then. Still it might be a good idea to do that.

I'll make a new tag for tracking binary file changes so I can verify that code and binaries don't get out of sync.

HellDragon-HK:
They are on my Google Drive, I can post the link here easy enough, here you go 
https://drive.google.com/open?id=1vBt0J76ByX2sOVR09lTxcKPWdstLQEGb

GreatEmerald:
Volume increasing by editing the sounds is a bad idea, that should be done in code (else you get clipping). But UT3 has a lot of special effects for sounds, which may need to be applied by editing them. Something to check...

HellDragon-HK:
I use mp3gain for volume editing which detects if its going to have clipping or not

GreatEmerald:
Yes, that's an OK solution too, but usually you can't increase it by much due to clipping anyway.

You can see how volume works here: http://greatemerald.xmpcommunity.com/uncodex/UT2004/Source_onslaughtbp/onsdualacsidegun.html#336 It's a parameter to the `PlaySound` function, which can be made into a defaultproperty (but is not always done so).

HellDragon-HK:
Missile load isn't a fire sound though it's something else, so how do those get handled?

HellDragon-HK:

PR #30, opened 2017-12-13
UT3 Scorpion Damage Phases Timing Closer to UT3

HellDragon-HK:
Smoke now starts at exactly the same health remaining as UT3 which is 192 health (if shot only by an enforcer from full 300)
Fire starts at 109 which is as close as I could get it, in UT3 it starts at 111


GreatEmerald:
You can use more numbers after the point. If `0.37` is not precise enough, use `0.373` or whatever.

Also again do not use a tab there, use four spaces.

HellDragon-HK:
Ok so keep tweaking the fire damage setting still, I'll look into it

HellDragon-HK:
Your estimate of 373 looks to probably be right, I can't get weapon damage to hit 111 but it catches fire now at 108 which is close enough

GreatEmerald:
OK, thanks.

PR #31, opened 2017-12-14
UT3 Manta Closer to UT3 Impact Damage

HellDragon-HK:
In UT3 the Manta (and actually most vehicles) takes little to no damage on impact, usually around 1 for a wall and 4 for an enemy vehicle or so,  while this value isn't quite perfect, it's close to what UT3 is with the Manta only taking damage while near full speed and only 1-3 usually for a wall.  It might be the damage is reduced too much with this but anything higher results in damage from barely any speed.

### DIFF
--- a/Classes/UT3Manta.uc
+++ b/Classes/UT3Manta.uc
@@ -1,7 +1,7 @@
 /*
- * Copyright © 2008 Wormbo
- * Copyright © 2012 100GPing100
- * Copyright © 2008, 2014 GreatEmerald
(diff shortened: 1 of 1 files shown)


GreatEmerald:
You can also use more numbers after the point too if you want to try and make it more precise.

HellDragon-HK:
I don't know how to test this further, it seems like it's handled very differently between games

HellDragon-HK:
Now included in the Manta Full Game Play Improvement request and not needed

PR #32, opened 2017-12-14
UT3 Manta Damage Phase Timing Closer to UT3

HellDragon-HK:
I think I got both Smoke and Fire phases almost identical to UT3 here as both start smoking at 128 health and catch fire at 73 health....well identical in remaining health but visually still not right by any means which is beyond my ability I think


GreatEmerald:
Yes, replicating the pieces falling off would be really quite difficult.
This looks OK, just convert your tabs to spaces.

HellDragon-HK:
Not the pieces, the location of the fire & smoke and needing to add sparks is what I meant, I really didn't expect UT2004 could do the pieces part which would be cool but doubt it can be done

HellDragon-HK:
Now included in the Full Game Play Improvement request so duplicate

PR #33, opened 2017-12-14
UT3 Hellbender Driver Turret Movement Closer to UT3

HellDragon-HK:
In UT3 on PC (not PS3 though) the turrets don't have instant turn and for the driver turret it can shoot straight up into the sky but has limited pitch on firing low, these values feel the closet to this


GreatEmerald:
But is the rotation speed fine with the default values?
Also, you need tabs to spaces again.

HellDragon-HK:
The speed is still an issue and I haven't figured out how to do that yet on anything but the Paladin, they do need to turn faster though yes

GreatEmerald:
OK, this is good enough. The speed can be looked into later.

PR #34, opened 2017-12-14
UT3 Hellbender No Impact Damage Discussion

HellDragon-HK:
Smoke starts at same health but on fire is off by 10 points of health, the damage per sec should be consistent across every vehicle but for some reason the UT3 Hellbender in UT2004 is not taking ANY fire damage

### DIFF
--- a/Classes/UT3Hellbender.uc
+++ b/Classes/UT3Hellbender.uc
@@ -420,6 +420,9 @@ defaultproperties
     IdleSound=Sound'UT3A_Vehicle_Hellbender.Sounds.A_Vehicle_Hellbender_EngineIdle01'
     StartUpSound=Sound'UT3A_Vehicle_Hellbender.Sounds.A_Vehicle_Hellbender_EngineStart01'
     ShutDownSound=Sound'UT3A_Vehicle_Hellbender.Sounds.A_Vehicle_Hellbender_EngineStop01'
+    DamagedEffectHealthSmokeFactor=0.65 //0.5
+	DamagedEffectHealthFireFactor=0.37 //0.25
+	DamagedEffectFireDamagePerSec=0.95 //0.75
     EntryRadius=300.000000
     TPCamWorldOffset=(Z=200.000000)
     MomentumMult=1.000000
--- a/Classes/UT3Hellbender.uc
+++ b/Classes/UT3Hellbender.uc
@@ -421,8 +421,8 @@ defaultproperties
     StartUpSound=Sound'UT3A_Vehicle_Hellbender.Sounds.A_Vehicle_Hellbender_EngineStart01'
     ShutDownSound=Sound'UT3A_Vehicle_Hellbender.Sounds.A_Vehicle_Hellbender_EngineStop01'
     DamagedEffectHealthSmokeFactor=0.65 //0.5
-	DamagedEffectHealthFireFactor=0.37 //0.25
-	DamagedEffectFireDamagePerSec=0.95 //0.75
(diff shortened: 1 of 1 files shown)


GreatEmerald:
Again you can make this more precise, again tabs to spaces. As for the fire damage, that ought to be controlled by the `DamagedEffectFireDamagePerSec` variable, yes... I'm not seeing any code that would suggest that the Hellbender behaves any differently from other vehicles here. The only thing that might affect that is https://github.com/GreatEmerald/UT3Vehicles/blob/master/Classes/UT3Hellbender.uc#L139 using `SVehicle` instead of `ONSVehicle`, which may have some native code to enable damage taking. I'd have to compare these functions to check what I changed there that required overriding the function in the first place...

HellDragon-HK:
I'm thinking it's the PRV code itself somewhere for not taking damage because I ran across another vehicle with the same issue, I need to check the stock Hellbender I think in-game for this

HellDragon-HK:
@GreatEmerald The stock Hellbender does burn so something is different in the UT3 mod code

GreatEmerald:
Yea, that must be it, then, that line I pointed to. You could test by temporarily commenting out that entire function.

HellDragon-HK:
Uh the Hellbender is now unusable without that, it just wheelies and won't move forward or back

GreatEmerald:
Did you test by commenting out the entire function? I mean the whole thing, not just the call to `Super`.

HellDragon-HK:
Do you mean adding // to lines 139-203?

GreatEmerald:
You can use `/*` to start a comment and `*/` to stop it, so you don't have to add `//` on every single line. But I actually mean the whole thing, from line 128 to 203.

HellDragon-HK:
So then 203 is the end then, ok I'll try it

HellDragon-HK:
@GreatEmerald That's even worse results, not only does it still act like a horse but it now has no textures or weapons

HellDragon-HK:
No longer needed as it's included in the Overall but we have discussion on what's going on with no impact damage so I'm leaving this open

GreatEmerald:
You must still be commenting it out incorrectly. You need to comment out the function definition, not the contents! The line `simulated function PostNetBeginPlay()` must be commented out too.

I looked at what the differences are between that function and vanilla functions and I cannot see any at a glance. There's no way it would have a big effect if you comment out the whole thing.

HellDragon-HK:
Nope not what's causing it, after doing that the Hellbender still looks normal but it comes with a floating stock Hellbender turret on the hood that doesn't move with it, other than that it's the same and no impact or fire damage

GreatEmerald:
That is very strange. Does the stock Hellbender get fire damage as it should?

HellDragon-HK:
HellDragon-HK commented 10 days ago
@GreatEmerald  The stock Hellbender does burn so something is different in the UT3 mod  code, probably related to whatever is disallowing collision damage to
--

GreatEmerald:
Oh, of course! It's damage inflicted by the vehicle unto itself. That's disallowed by: https://github.com/GreatEmerald/UT3Vehicles/blob/master/Classes/UT3Hellbender.uc#L58
Try replacing the `if` statement by:
```UnrealScript
if (InstigatedBy == self && ClassIsChildOf(DamageType, class'DamTypeSkyMine'))
    return;
```

HellDragon-HK:
Hey that did the trick and the Hellbender is now taking damage from walls (a lot of damage actually, time to find an ImpactDamageMult value again) as well as fire damage, I'll add it to the overall request immediately and add the impact mult when I figure that out

PR #35, opened 2017-12-14
UT3 Hellbender Rear Turret Closer to UT3

HellDragon-HK:
In UT3 the Hellbender turrets don't turn instantly on PC and I've gotten consistent 85 damage to every vehicle I've shot with the rear turret

### DIFF
--- a/Classes/UT3HellbenderRearGun.uc
+++ b/Classes/UT3HellbenderRearGun.uc
@@ -156,8 +156,8 @@ defaultproperties
     FireSoundClass=sound'UT3A_Vehicle_Hellbender.Sounds.A_Vehicle_Hellbender_TurretFire01'
     //BeamEffectClass=class'ShockBeamEffect'//'ONSChargeBeamEffect'
     FireInterval=0.5
-    DamageMin=180
-    DamageMax=180
+    DamageMin=85 //180
+    DamageMax=85 //180
     Momentum=75000
     MaxHoldTime=3.0
     //MinDamageScale=1.0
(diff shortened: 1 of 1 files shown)


GreatEmerald:
Again this is just scaling damage against vehicles: http://greatemerald.xmpcommunity.com/uncodex/UT3-15/Source_utgamecontent/utdmgtype_hellbenderprimary.html

HellDragon-HK:
So look at the LeviRocketTurret code you mean again?

GreatEmerald:
Yeap, same principle.

HellDragon-HK:
Is 120 the base damage because that's what I'm seeing in the UT3 code here http://greatemerald.xmpcommunity.com/uncodex/UT3-15/utgamecontent/utvweap_hellbenderprimary.html

so that is being scaled by the .85?

GreatEmerald:
Correct, it's 120 damage and it gets scaled by 0.85 due to the damage type. This pull is fine now, but the damage scaling for vehicles still needs to be implemented.

HellDragon-HK:
I thought I did add the scaling, what do you mean?

GreatEmerald:
You need to add the DamageType .uc and set `MyDamageType` here.

HellDragon-HK:
Wait you merged it before having the DamageType added?

GreatEmerald:
Yes, as you can see. You didn't set `MyDamageType` so this doesn't cause a problem; it's simply just as inaccurate as it was before. Please open a new pull request with the `DamageType` added and applied.

PR #36, opened 2017-12-14
UT3 Hellbender Missing Sounds Fix

HellDragon-HK:
This adds in the missing sounds the Hellbender has in UT3, according to the editor it uses only Scorpion Collide 03 & 04 for it's impacts
NOTE: Updated Scorpion UAX is needed


HellDragon-HK:
Included in Overall now so not needed

PR #37, opened 2017-12-14
UT3 Hellbender Drawscale Discussion

HellDragon-HK:
First off I was wrong on DrawScale as my preferred character model is apparently an inch or more taller than a normal UT2004 model so using a normal UT2004 model a DrawScale of .9 looks closer, I've done my best to tweak the coronas to work with the smaller size.

Ground Speed according to UnCode is 800 but the max that can be achieved is 1050, TransRatio (assuming I understand it) has been increased to give acceleration closer to UT3 which is quite fast and probably still faster (assuming this is what TransRatio is)

MaxBrakeTorque of at least 30 is needed to come close to replicating how easily the Hellbender in UT3 can stop but in UT2004 it causes it to be lifted off the back wheels to the point it's bottom up so can't do that yet.

Whew a lot of testing and tweaking today....it's now 1:28 AM and I will probably be in bed until somewhere after 9 AM at least so don't expect my usual snappy replies....good night :)

### DIFF
--- a/Classes/UT3Hellbender.uc
+++ b/Classes/UT3Hellbender.uc
@@ -432,7 +432,7 @@ defaultproperties
     SoundVolume=255
     TransRatio=0.18 //0.11
     
-	DrawScale=0.9
+    DrawScale=0.9
     
     CollisionRadius=219
     HeadlightCoronaOffset(0)=(X=68.5,Y=24.5,Z=47.5) //(X=77.5,Y=27.5,Z=52.5)
@@ -442,10 +442,10 @@ defaultproperties
     HeadlightCoronaMaterial=Material'EmitterTextures.Flares.EFlareOY'
     HeadlightCoronaMaxSize=82
 
-	HeadlightProjectorOffset=(X=78.0,Y=0,Z=50.5) //(X=82.5,Y=0,Z=55.5)
-	HeadlightProjectorRotation=(Yaw=0,Pitch=-1000,Roll=0)
-	HeadlightProjectorMaterial=Texture'VMVehicles-TX.NewPRVGroup.PRVProjector'
-	HeadlightProjectorScale=0.40 //0.65
+    HeadlightProjectorOffset=(X=78.0,Y=0,Z=50.5) //(X=82.5,Y=0,Z=55.5)
+    HeadlightProjectorRotation=(Yaw=0,Pitch=-1000,Roll=0)
+    HeadlightProjectorMaterial=Texture'VMVehicles-TX.NewPRVGroup.PRVProjector'
+    HeadlightProjectorScale=0.40 //0.65
     //HeadlightProjectorMaterial=None
 
     BrakeLightOffset(0)=(X=-124.5,Y=36.5,Z=58) //(X=-137.5,Y=42.5,Z=64)
--- a/Classes/UT3Hellbender.uc
+++ b/Classes/UT3Hellbender.uc
@@ -428,20 +428,27 @@ defaultproperties
     VehiclePositionString="in a Hellbender"
     VehicleNameString="UT3 Hellbender"
     HornSounds(0)=Sound'UT3A_Vehicle_Hellbender.Sounds.A_Vehicle_Hellbender_Horn01'
-    GroundSpeed=700.000000
+    GroundSpeed=800.000000 //700
     SoundVolume=255
-
-    DrawScale=1.0
+    TransRatio=0.18 //0.11
+    
+	DrawScale=0.9
+    
     CollisionRadius=219
-    HeadlightCoronaOffset(0)=(X=77.5,Y=27.5,Z=52.5)
-    HeadlightCoronaOffset(1)=(X=77.5,Y=-27.5,Z=52.5)
-    HeadlightCoronaOffset(2)=(X=77.5,Y=25,Z=41)
-    HeadlightCoronaOffset(3)=(X=77.5,Y=-25,Z=41)
+    HeadlightCoronaOffset(0)=(X=68.5,Y=24.5,Z=47.5) //(X=77.5,Y=27.5,Z=52.5)
+    HeadlightCoronaOffset(1)=(X=68.5,Y=-24.5,Z=47.5)
+    HeadlightCoronaOffset(2)=(X=68.5,Y=22,Z=38) //(X=77.5,Y=-25,Z=41)
+    HeadlightCoronaOffset(3)=(X=68.5,Y=-22,Z=38)
     HeadlightCoronaMaterial=Material'EmitterTextures.Flares.EFlareOY'
-    HeadlightCoronaMaxSize=94
-    HeadlightProjectorMaterial=None
+    HeadlightCoronaMaxSize=82
 
-    BrakeLightOffset(0)=(X=-137.5,Y=42.5,Z=64)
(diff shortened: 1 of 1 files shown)


GreatEmerald:
`TransRatio` has something to do with switching gears. I'm not sure what exactly, since it's implemented in native code (possibly Karma) that is not public... But if it helps, might as well, I guess.

For the DrawScale changes, I'm still not a fan of changing it without some good numerical reason as to why it should be like that... It doesn't match UT3, and supposedly it's elongated, which may be because of some porting issues, of course, but still.

If the `MaxSpeed` is different, why didn't you change it as well? (I'm guessing the difference between them is that you can't achieve `MaxSpeed` yourself, but if some external force accelerates you to it, you keep it instead of slowing down.)

And this needs tabs to spaces too.

Classes/UT3Hellbender.uc
```diff
@@ -428,20 +428,27 @@ defaultproperties
...
+    HeadlightCoronaOffset(1)=(X=68.5,Y=-24.5,Z=47.5)
+    HeadlightCoronaOffset(2)=(X=68.5,Y=22,Z=38) //(X=77.5,Y=-25,Z=41)
+    HeadlightCoronaOffset(3)=(X=68.5,Y=-22,Z=38)
     HeadlightCoronaMaterial=Material'EmitterTextures.Flares.EFlareOY'
-    HeadlightCoronaMaxSize=94
-    HeadlightProjectorMaterial=None
+    HeadlightCoronaMaxSize=82
+
+	HeadlightProjectorOffset=(X=78.0,Y=0,Z=50.5) //(X=82.5,Y=0,Z=55.5)
+	HeadlightProjectorRotation=(Yaw=0,Pitch=-1000,Roll=0)
```
GreatEmerald: Also, this is clearly not exact: rotators in the Unreal Engine are in RUUs that have a range of 0-65535. So if it is close to -1000, then it should be set to -1024 (that's a round number).

HellDragon-HK:
The original 1.0 Drawscale conclusion was because I was measuring with a much taller character than stock UT2004 characters, measuring with a stock UT2004 character and it didn't even come close to matching up with UT3 from any angle

MaxSpeed doesn't exist in UT2004 code that I can see so how do we set it?

I didn't change rotation, that's what it is by default

GreatEmerald:
It's part of `KarmaParamsRBFull`, variable `kMaxSpeed`: http://greatemerald.xmpcommunity.com/uncodex/UT2004/Source_onslaught/onsprv.html#423

If those are defaults, don't paste them in, they are already defaults.

HellDragon-HK:
Ok just need to remove the rotation part then

HellDragon-HK:
Adding the KMaxSpeed really made a difference, it really does feel like the UT3 Hellbender in that regard now (to me at least)

HellDragon-HK:
@GreatEmerald Can you please look at these couple of screenshots and tell me what it is because it's driving me crazy, is it wheel size, is it body size, is it something to do with animations or unknown factor...what? They look the same in the editors to me so I'm really at a loss.
https://drive.google.com/open?id=18PW5_BtnugwypcyzAwv0s9j7PSyz67_h

Unre-Alex:
I took a ruler and started to measure both the Hellbender and the Scorpion's width (by standing at the top), and got the same differences between games when I scaled down the Hellbender to 0.95. BUT, assuming that the Manta in UT2004 is the right scale (because of the sitting player's height and the drawscale of 1.0) then the Hellbender would be 1.0, the Scorpion 1.05, the SPMA 1.1, the Paladin and the Goliath 1.2. I couldn't jump onto the flying ones.

HellDragon-HK:
I'm confused there, do you mean that all of the vehicles including the Scorpion need to be sized up or that the Hellbender needs to be 0.95? which I think I've got it at 0.9 right now.  We know the Goliath needs to be scaled up but it seems impossible :(

Unre-Alex:
That's the question. With the Manta and The Hellbender being 1.0 that makes them 0.15 smaller than in UT3 (shouldn't be 0.25?). 0.9/5 looks small to me. Do you use "behindview 1" command in UT3 to compare them with a character or some 3rd person view mutator?

HellDragon-HK:
ActionCam mutator, I don't use console commands often because I don't know them and I usually do with 3rd person mutators so hadn't even considered a console command

GreatEmerald:
Very interesting, thanks for the nice comparison screenshots. This shows that indeed there is some bug in the code.

Clearly the difference between the games is the suspension: it is much closer to the body of the vehicle in UT3, whereas it gets real far in UT2004. The odd thing is that it is not the case in the editor.

I think the way to fix it is to experiment with bone names and offsets, see https://github.com/GreatEmerald/UT3Vehicles/blob/master/Classes/UT3Hellbender.uc#L363: the `SupportBoneName` probably should be set to the suspension bones so as not to distort it. You can also experiment with `BoneOffset` and setting different bone names for both the wheel and the suspension (see the bone view in the editor to get the names).

HellDragon-HK:
Can you explain how those function so I have a better idea of what I'm messing with, remember I'm new to all this and don't know what you're talking about most of the time GE, and you're welcome glad to finally put the issue to rest on if it's wheel size, elongated, or something else.

GreatEmerald:
I'm not too sure myself what those do, hence why I'm suggesting that you play around with the values and see what happens. I know that `BoneName` is the name of the bone that gets rotated whenever the vehicle moves (which is why the Hellbender and SPMA have the problem of fenders turning together with the wheels: the fender bone is attached to the wheel bone). I don't really know what `BoneOffset` is about, but it's some offset from the bone itself (perhaps to make it rotate along a different point in space). This might be the reason why the wheels are so far off from how they look in the editor, though, so you should test what it really does.

`SupportBoneName` sounds interesting, this should be the suspension that moves along with the wheel when it bounces. It might indeed help in this case as well, especially to make it close to how it is in UT3. This might need to be set to the `*_Control` bones, it's mentioned in UT3 source. Related parameters are `SupportBoneAxis` (on which axis it moves as it bounces), `SuspensionTravel` (how much is the suspension allowed to move when bouncing), `SuspensionOffset` (I guess this is for when the bone is not quite in the right place) and `SupportPivotDistance` (apparently something for calculating exactly how it's supposed to bounce).

For reference, there's a bunch of parameters in UT3 that are different for the Hellbender: http://greatemerald.xmpcommunity.com/uncodex/UT3-15/Source_utgame/utvehiclehellbenderwheel.html for overall changes and http://greatemerald.xmpcommunity.com/uncodex/UT3-15/Source_utgame/utvehicle_hellbender.html#110 for per-wheel changes.

It's still strange that it's so far away from what it looks like in the editor with the current code, though. I'm guessing it might be the `BoneOffset` that's causing all the problems here.

HellDragon-HK:
I fixed it already with the WheelsSuspensionOffset which apparently says how high or low it is and Alex confirmed it works, did I forget to add it to the request oops

"I know that BoneName is the name of the bone that gets rotated whenever the vehicle moves (which is why the Hellbender and SPMA have the problem of fenders turning together with the wheels"

How does it work for the stock Paladin then, it's fenders are actually part of the wheel from what I can tell and they don't turn with the wheels :S

GreatEmerald:
Oh, that's nice to hear! Then open a new pull request or include that into the overall improvements.

HellDragon-HK:
It already is included in the overall

PR #38, opened 2017-12-14
UT3 Raptor Damage Phases Closer to UT3

HellDragon-HK:
Smoke starts at exactly 192 hp like it does in UT3 but On Fire phase is 5 points off which isn't too horrible I guess


GreatEmerald:
Tabs and precision here too.

HellDragon-HK:
Don't think I can get it closer, it jumps by 10 points with almost any value (for the Hellbender it did)

HellDragon-HK:

PR #39, opened 2017-12-14
UT3 Manta Gun Barrel Movement

HellDragon-HK:
Add this in so the Manta's guns at least move up and down, anything else is going to have to be handled by you or 100Gping100

### DIFF
--- a/Classes/UT3Manta.uc
+++ b/Classes/UT3Manta.uc
@@ -223,9 +223,10 @@ simulated function CheckJumpDuck()
 simulated function EmeraldTick(float DeltaTime)
 {
 Super.Tick(DeltaTime);
+Guns();
 if (!bHoldingDuck && DuckEffect!=None) {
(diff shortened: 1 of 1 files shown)


HellDragon-HK:
Oops forgot something,

simulated function EmeraldTick(float DeltaTime)
{
Super.Tick(DeltaTime);
Guns();
if (!bHoldingDuck && DuckEffect!=None) {
    DuckEffect.Destroy();   
	}
}

GreatEmerald:
Hah. You'd need to add those changes to the pull request.

Also, if there's a way to destroy the effect on an event rather than checking if it should be destroyed on each frame, that would be better. Not sure if there's a function for that, though.

HellDragon-HK:
All I added was the Guns() the rest of that is your handiwork :P

HellDragon-HK:
Included in overall, no longer needed

PR #40, opened 2017-12-14
UT3 Raptor Plasma Fire Location Change

HellDragon-HK:
After a lot of tweaking this is currently the best I've been able to get the plasma to actually come from the gun barrels, I think all that is needed now is to move it downward but I haven't been able to figure out how and using the full (XYZ) doesn't seem to do anything.


HellDragon-HK:
Actually after closer examination this might be better
    DualFireOffset = 37.0 //15.0

GreatEmerald:
In that case, update the value.

HellDragon-HK:
No problem, done and good morning

GreatEmerald:
OK, looks good.

PR #43, opened 2017-12-14
UT3 Raptor Impact Damage Closer to UT3

HellDragon-HK:
In UT3 the Raptor (and all vehicles in general) takes a lot less impact damage from hitting objects such as walls.   This value seems to get it closer to that, in UT3 skirting the ground doesn't damage the Raptor and now it won't happen in UT2004 either.  Flying almost full speed into a wall will still do 1-4 damage though (especially sideways it seemed) which is about what UT3 is.


GreatEmerald:
Looks good to me.

PR #44, opened 2017-12-15
UT3 Cicada Damage Phases Closer to UT3

HellDragon-HK:
Smoke phase starts at exactly the same health as in UT3 but On Fire is again 10 points late


GreatEmerald:
Again tabs and numbers after the decimal point.

HellDragon-HK:
Ok let me try to get the fire timing a little better first

HellDragon-HK:
Included in overall request and no longer needed

PR #45, opened 2017-12-15
UT3 Hellfire SPMA Sound Update

HellDragon-HK:
This adds the missing collision (which according to UT3 Editor is Goliath collide sound) and explosion sounds (uax needs to be updated for explosion sound, it can be found on SPMA issues comments area)


HellDragon-HK:
No longer needed, included in Overall

PR #46, opened 2017-12-15
UT3 Hellfire SPMA Side Gun Sound Update

HellDragon-HK:
This adds the sound for the turret rotate, the InstantRotation=False is needed for it to work, and in UT3 the turret turns fast but not instantly anyway if you aim left then right quickly you can see it takes it a little time to turn


GreatEmerald:
I think all sound update pulls can be merged into one.
As for the rotation, check the speeds.

HellDragon-HK:
So far they have all been ignoring RoationsPerSecond, what else can I try because it does need an increase no doubt about that

Unre-Alex:
I don't hear that sound in UT3, might be seen as an improvement but the turret doesn't turn with the vehicle (the same in UT3) and you hear that sound all the time.

HellDragon-HK:
I have to agree with it being slightly annoying (maybe why it was unused although I could swear I've heard it in-game and think I have a checkbox for it even) but we can solve that by reducing volume maybe?

HellDragon-HK:
Redone

PR #47, opened 2017-12-15
UT3 Hellfire SPMA Cannon Sound Update

HellDragon-HK:
You had the UT3 Cannon Rotate sound in the package but forgot to use it apparently?


GreatEmerald:
Oh, that's useful, thanks for noticing!

PR #48, opened 2017-12-15
UT3 Hellfire SPMA Camera Size Fix

HellDragon-HK:
In UT3 the camera is fairly small, this probably isn't exact but it shouldn't be far off, also according to the sound files there should be this ambient sound but I haven't been able to get it to work so it seems to do nothing but I believe this is where it goes.

The explosion size is still large, I haven't found how to scale it down yet but in UT3 it's about the same size as the camera.

### DIFF
--- a/Classes/UT3HellfireSPMACamera.uc
+++ b/Classes/UT3HellfireSPMACamera.uc
@@ -333,7 +333,8 @@ defaultproperties
     CVScale = 1.0
     MaxTargetRange = 10240.0
     Speed = 4000.0
-
+    DrawScale=0.3
+    AmbientSound      = Sound'UT3SPMA.SPMACameraAmbient'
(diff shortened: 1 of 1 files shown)


GreatEmerald:
The explosion size is `DamageRadius`, if you want to adjust that.

HellDragon-HK:
Can you point me to the class I need for that, there are so many for the SPMA I can't tell what is what, and I don't want to mess with the actual artillery fire shell radius

GreatEmerald:
Oh, you mean the emitter, OK. Yea, you can't easily scale those I'm pretty sure.

HellDragon-HK:
So leave that for later and call it a done for now, probably needs a new visual suited for it instead of trying to scale the old anyway right?

GreatEmerald:
Exactly.

PR #49, opened 2017-12-15
UT3 Hellfire SPMA Headlights Adjustment

HellDragon-HK:
We've got a slight problem with the headlights, they are meant to come from the wheel plates which turn and change position while moving forward or backward causing floating headlights....I'm guessing there isn't a way to fix this so the question is which do we want, headlights that are lined up when the vehicle is sitting still (probably best considering SPMA is usually deployed when you see it) or do we want them to line up while moving forward?  These values are for sitting still, I haven't come up with ones for moving yet but will if I need to.

### DIFF
--- a/Classes/UT3HellfireSPMA.uc
+++ b/Classes/UT3HellfireSPMA.uc
@@ -929,8 +929,8 @@ defaultproperties
     DriverDamageMult = 0.0
     TreadVelocityScale = 30.0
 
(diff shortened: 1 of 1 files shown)


GreatEmerald:
Like I said in the other thread, I'd vote for neither. It interferes with immersion if it's flating in either case.

HellDragon-HK:
Ok then remove this topic

HellDragon-HK:
No longer needed

PR #50, opened 2017-12-16
UT3 Hellfire SPMA Handling Closer to UT3

HellDragon-HK:
Some tweaks to make it handle more like UT3, I managed to get the speed to feel exactly like UT3 which I'm not sure we want because it's VERY VERY slow and UT2004 maps are (or seem) larger than UT3, and while I haven't tried it, it would probably take more than 3 minutes to cross Torlan with this speed.

In UT3 you can't do the handbrake turn, despite setting it to false it still happens in UT2004 and I haven't found a way to stop it yet.

### DIFF
--- a/Classes/UT3HellfireSPMA.uc
+++ b/Classes/UT3HellfireSPMA.uc
@@ -928,6 +928,12 @@ defaultproperties
     bDrawDriverInTP = false
     DriverDamageMult = 0.0
     TreadVelocityScale = 30.0
+    SteerSpeed=40 //110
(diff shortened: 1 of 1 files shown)


GreatEmerald:
Tabs to spaces needed.
The handbrake change might interfere with deployment. Have you checked that it works as expected?

HellDragon-HK:
I did actually yes it still worked but I'll double check to be sure

HellDragon-HK:
Redone in an Overall request, no longer needed

PR #51, opened 2017-12-16
UT3 Hellfire SPMA Damage Phases Closer to UT3

HellDragon-HK:
Smoke started at exactly 518 hp on both games but on fire started at 313 in UT3 and 296 in UT2004....and like the Hellbender fire damage over time doesn't seem to be happening no matter what value is used.


GreatEmerald:
Tabs and precision again.

Hm, strange that damage over time doesn't happen, considering that in this case there seems to be no calls to `Super(SVehicle)` unlike with the Hellbender... But with native code it's hard to say what's going on here.

HellDragon-HK:
Format fixed and a value closer on timing, it started it's on fire phase at 308 which is within the 6 damage the enforcer is doing so it might actually be exact now

HellDragon-HK:

PR #54, opened 2017-12-16
UT3 Paladin Wheel Movement Fix

HellDragon-HK:
In UT3 all wheels move, this should allow that


100GPing100:
Tested and approved. The reason this was not done before was for aesthetics; less wheel going through the fenders.

I might have a solution to make the fenders move with the wheels, it all depends on how the mesh is mapped to the bones.

HellDragon-HK:
I looked at the bones and it looks like the Tires and Fenders are literally on top of each other as the names are on top of one another by the look of it....could be wrong as I'm still learning how things work :)

GreatEmerald:
Superseded.

PR #55, opened 2017-12-16
UT3 Paladin Damage Phases Closer to UT3

HellDragon-HK:
As usual the smoke phase is right on target, fire is still a little late but for now it's good


100GPing100:
Should be accepted once you can get the fire just right ;)

GreatEmerald:
Superseded.

PR #56, opened 2017-12-16
UT3 Paladin Cannon Movement Closer to UT3

HellDragon-HK:
Not perfect but closer than it was, the turret rotation speed probably needs to be more still but this should do for now


100GPing100:
Tested this and approved.

HellDragon-HK:
@100GPing100 Thanks

PR #59, opened 2017-12-18
UT3 Scorpion 32 Bit Sound Difference Fix

HellDragon-HK:
I don't know what these numbers do but I do know they are the cause of the sound not being right in 32 bit, without this number it sounds right in both 32 & 64 bit now

### DIFF
--- a/Classes/UT3Scorpion.uc
+++ b/Classes/UT3Scorpion.uc
@@ -239,7 +239,7 @@ function Boost()
     {
         //log("UT3: Boosting!");
         BoostRechargeCounter=0;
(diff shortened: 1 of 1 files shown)


HellDragon-HK:

GreatEmerald:
Hm, this is quite interesting to know. The parameters to `PlaySound` are:
```UnrealScript
native(264) final function PlaySound ( sound	Sound, optional ESoundSlot Slot, optional float	Volume, optional bool	bNoOverride, optional float	Radius, optional float	Pitch, optional bool	Attenuate )
```
So it's the pitch. I guess 64-bit doesn't support pitch changes.

PR #60, opened 2017-12-18
UT3 Manta Aim Closer to UT3

HellDragon-HK:
In UT3 the Manta has better aim ability than the UT2004 Manta by quite a lot so this is an attempt to fix that.


GreatEmerald:
Looks good, thanks.

PR #61, opened 2017-12-19
UT3 Manta Full Game Play Improvement Closer to UT3

HellDragon-HK:
Let's try this again
Tweaks include
-GroundSpeed
-MaxSpeed
-Turning (unchanged from last I don't think, just moved to Movement section)
-Roll (just moved to movement)
-Pitch (just moved to movement)
-Strafe Increase
-Hover Height
-Hover Bounce
-Hover Stability (this is actually what we need to reduce to get roll when turning but it causes alt fire to go straight to the ground and scrape it making movement difficult)
-Damage Phases closer to UT3 (At what health a vehicle starts smoking, becomes on fire, and how fast fire damages the vehicle)
-Momentum (How much the vehicle gets thrown around by projectiles, probably not perfect yet but seems close at this time, this will probably be a hard one because we have to consider all other vehicles and weapons at the same time)
-Exit Positions
-HeadlightProjector (Needs new texture for single round circle, and at this time it's using the Scorpion's quad circle)

### DIFF
--- a/Classes/UT3Manta.uc
+++ b/Classes/UT3Manta.uc
@@ -322,6 +322,9 @@ defaultproperties
     ShutDownSound = Sound'UT3A_Vehicle_Manta.Sounds.A_Vehicle_Manta_Stop01';
     JumpSound = Sound'UT3A_Vehicle_Manta.Sounds.A_Vehicle_Manta_Jump';
     DuckSound = Sound'UT3A_Vehicle_Manta.Sounds.A_Vehicle_Manta_Crouch';
+    DamagedEffectHealthSmokeFactor=0.65 //0.5
+    DamagedEffectHealthFireFactor=0.40 //0.25 //.373 is what I had, I think 0.4 will be too much but we'll se
+    DamagedEffectFireDamagePerSec=2.0  //0.75
     ImpactDamageSounds(0) = Sound'UT3A_Vehicle_Manta.Sounds.A_Vehicle_Manta_Collide01';
     ImpactDamageSounds(1) = Sound'UT3A_Vehicle_Manta.Sounds.A_Vehicle_Manta_Collide02';
     ImpactDamageSounds(2) = Sound'UT3A_Vehicle_Manta.Sounds.A_Vehicle_Manta_Collide01';
--- a/Classes/UT3Manta.uc
+++ b/Classes/UT3Manta.uc
@@ -2,6 +2,7 @@
  * Copyright © 2008 Wormbo
  * Copyright © 2012 100GPing100
  * Copyright © 2008, 2014 GreatEmerald
+ * Copyright © 2018 HellDragon
  *
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions are
@@ -42,8 +43,8 @@
 class UT3Manta extends ONSHoverBike;
 
 var Emitter DuckEffect;
-Var rotator ThighDriveL,ThighDriveR;
-var bool bAttachedDriver;
+//Var rotator ThighDriveL,ThighDriveR;
+//var bool bAttachedDriver;
 
 //===============
 // @100GPing100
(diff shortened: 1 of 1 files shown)


GreatEmerald:
Thanks. A few comments:
```UnrealScript
Var rotator ThighDriveL,ThighDriveR;
```
These could be local variables, since you're not using them anywhere else in the code except that one function.

Next, `bAttachedDriver` seems pointless, it's unused.

Lastly:
> Damage Phases closer to UT3

This doesn't seem to be present.

HellDragon-HK:
It won't compile without the bAttachedDriver, and I just learned how to use the rotator to fix driving positions so I don't understand what you mean by local variables, how else can it be done?

I missed the damage phases? Huh must have thought they were already in the class or something, I'll fix that.

Classes/UT3Manta.uc
```diff
@@ -42,6 +42,8 @@
 class UT3Manta extends ONSHoverBike;
 
 var Emitter DuckEffect;
+Var rotator ThighDriveL,ThighDriveR;
```
GreatEmerald: This can be put between lines 220 and 221 as
```UnrealScript
local rotator ThighDriveL,ThighDriveR;
```
Alternatively, you can remove lines 224 and 226 and instead add to `defaultproperties`:
```UnrealScript
ThighDriveL = (Pitch=1800)
ThighDriveR = (Pitch=-1800)
```

Classes/UT3Manta.uc
```diff
@@ -42,6 +42,8 @@
...
+var bool bAttachedDriver;
```
GreatEmerald: You should remove this, as well as...

Classes/UT3Manta.uc
```diff
@@ -214,6 +216,38 @@ if (!bHoldingDuck && DuckEffect!=None) {
     }
 }
 
+simulated function AttachDriver(Pawn P)
+{
+    super.AttachDriver(P);
+    bAttachedDriver=true;
```
GreatEmerald: ...this...

Classes/UT3Manta.uc
```diff
@@ -214,6 +216,38 @@ if (!bHoldingDuck && DuckEffect!=None) {
...
+    P.SetBoneRotation('Bip01 L UpperArm');
+    P.SetBoneRotation('Bip01 R UpperArm');
+    P.SetBoneRotation('Bip01 L ForeArm');
+    P.SetBoneRotation('Bip01 R ForeArm');
+    P.SetBoneRotation('Bip01 L Thigh');
+    P.SetBoneRotation('Bip01 R Thigh');
+    P.SetBoneRotation('Bip01 L Calf');
+    P.SetBoneRotation('Bip01 R Calf');
+    
+    bAttachedDriver=false;
```
GreatEmerald: ...and this.

HellDragon-HK:
C:\Unreal Anthology\UT2004\UT3Vehicles\Classes\UT3Manta.uc(220) : Error, 'Var' is not allowed here
UT3Vehicles.UT3Manta: Unknown property in defaults:     ThighDriveL.Pitch=1800;
UT3Vehicles.UT3Manta: Unknown property in defaults:     ThighDriveR.Pitch=-1800;

GreatEmerald:
I said `local`, not var; and `ThighDriveL=(Pitch=1800)`, not ThighDriveL.Pitch=1800.

HellDragon-HK:
Ah ok sorry my bad, let me give it another shot and see what happens

HellDragon-HK:
C:\Unreal Anthology\UT2004\UT3Vehicles\Classes\UT3Manta.uc(222) : Error, 'Local' is not allowed here
Compile aborted due to errors.

GreatEmerald:
Don't do both, do either one or the other. Either local or put into defaultproperties.

HellDragon-HK:
Putting into default properties and commenting out all of AttachDriver (but not Detach I assume) doesn't work either

UT3Vehicles.UT3Manta: Unknown property in defaults:     ThighDriveL = (Pitch=1800)
UT3Vehicles.UT3Manta: Unknown property in defaults:     ThighDriveR = (Pitch=-1800)

GreatEmerald:
Put the first option into the pull request, I'll see what's going wrong there.

HellDragon-HK:
It finally did compile, no idea what I did different

GreatEmerald:
Yeap, that looks fine, now just remove the unneeded commented out parts.

PR #57, opened 2017-12-16
UT3 Paladin Handling Tweaks to be Closer to UT3

HellDragon-HK:
Some tweaks to handling to be closer to UT3

PR #53, opened 2017-12-16
UT3 Paladin Explosion Sound Correction

HellDragon-HK:
Explode01 is the Combo Fire sound and not for death explosion according to UT3 Editor

PR #52, opened 2017-12-16
UT3 Paladin Combo Fire Sound Fix

HellDragon-HK:
According to the UT3 editor this explosion is the combo explosion sound and not a death explosion sound.

PR #62, opened 2017-12-20
UT3 Paladin Fixes and Tweaks

100GPing100:
Deprecates pulls: #52 #53 #54 #55 #56 #57 

Takes care of all of @HellDragon-HK 's changes.
Fixes wrong sounds properly.


Classes/UT3Paladin.uc
```diff
@@ -1,7 +1,7 @@
 /*
- * Copyright � 2008-2009 Wormbo
- * Copyright � 2008-2009, 2014 GreatEmerald
- * Copyright � 2012 100GPing100
+ * Copyright © 2008-2009 Wormbo
+ * Copyright © 2008-2009, 2014 GreatEmerald
+ * Copyright © 2012-2017 Luís 'zeluisping' Guimarães <zeluis.100@gmail.com>
```
GreatEmerald: From what I recall, copyright law requires you to list years when work has been done on a file. Since there was no work done in between 2012 and 2017, these should all be `2012, 2017` instead of `2012-2017`. Also, in the cases where HellDragon found values, please add his copyright as well.

Classes/UT3Paladin.uc
```diff
@@ -49,181 +49,225 @@
...
+        KInertiaTensor(4)=0.0
+        KInertiaTensor(5)=3.0
+        KCOMOffset=(X=-0.25,Y=0.0,Z=-1.35)
+        bDestroyOnWorldPenetrate=True
+        bDoSafetime=True
+        Name="KParams0"
+    End Object
+    KParams=KarmaParams'KParams0'
+    
+    HeadlightCoronaOffset=()
```
GreatEmerald: This seems unnecessary, as far as I can see there are no headlight coronas in ONSShockTank to begin with.

Classes/UT3PaladinCannon.uc
```diff
@@ -163,27 +162,24 @@ function rotator SmoothRotate(float YawDelta, float PitchDelta, rotator CurrentR
...
-    //RotateSound=sound'ONSBPSounds.ShockTank.TurretHorizontal'
+    Mesh=SkeletalMesh'UT3VH_Paladin_Anims.PaladinWeaponOnly';
+    RedSkin=Shader'UT3PaladinTex.Paladin.PaladinSkin';
+    BlueSkin=Shader'UT3PaladinTex.Paladin.PaladinSkinBlue';
 
+    FireSoundClass=Sound'UT3A_Vehicle_Paladin.Sounds.A_Vehicle_Paladin_Fire01';
+    FireImpact=Sound'UT3A_Vehicle_Paladin.Sounds.A_Vehicle_Paladin_FireImpact01';
+    //RotateSound=sound'ONSBPSounds.ShockTank.TurretHorizontal'
+    RotationsPerSecond=0.68 //0.18
+	PitchUpLimit=9900
```
GreatEmerald: Please convert tab to spaces

Classes/UT3PaladinShield.uc
```diff
@@ -1,9 +1,43 @@
-//============================================================
-// UT3 Paladin Shield
-// Copyright (c) 2012, Jos� Lu�s '100GPing100'
-// Copyright (c) 2014, GreatEmerald
-// Contact: zeluis.100@gmail.com
-//============================================================
+/*
+ * Copyright © 2014 GreatEmerald
+ * Copyright © 2012-2017 Luís 'zeluisping' Guimarães <zeluis.100@gmail.com>
```
GreatEmerald: Thanks for updating the header, but if nothing else changed in the file, your copyright should state only the year 2012.

GreatEmerald:
@HellDragon-HK What would you like your copyright line to say? Also feel free to add yourself to copyright headers when sending pull requests.

HellDragon-HK:
I'm just trying to help so doesn't really matter to me much, btw I could have sworn there was a sound change in PaladinShield or something I did change something...let me look

GreatEmerald:
@HellDragon-HK Sure, but without you we wouldn't know all the values. You're doing most of the work here, so you should be recognised for it!

HellDragon-HK:
Thank you but it's all trial and error, I'm not a real coder but go ahead and just put whatever works for you and I'll start adding it to any of the requests I do I guess if that's what you need

HellDragon-HK:
Aw you want to remove the headlight, stock Paladin didn't have a spot for it but the UT3 has a great spot with the red bar light area

100GPing100:
@HellDragon-HK No no, the headlight is still there, we only removed `HeadlightCoronaOffset=()` which clears any previous headlights from the parent class, since the parent class has no headlights, then there is no need to clear it.

HellDragon-HK:
Ah ok cool I didn't know that, guess that's how the Cicada only has 1 now instead of 2

GreatEmerald:
Thanks!

PR #63, opened 2017-12-22
UT3 Cicada Overall Improvement

HellDragon-HK:
This is everything I've changed with the Cicada to get it closer to UT3 and add UT2004 only features which includes
-Speed (KMaxSpeed is already the same max speed as UT3 by default so wasn't needed)
-Handling (Still stops instantly with forward movement and I haven't figured that out yet if it's even possible to change, it also has the same turning issues stock UT2004 Cicada has which I can't seem to figure out either, other than that it should feel a little closer to UT3 now)
-Damage Phases more in-line with UT3 on when smoke and fire occur
-Fire Damage Rate
-Exit Positions closer to what UT3 has
-Trail Effect placement
-How much the Cicada is thrown around by weapon fire (might still need adjusting)
-Headlights adjustment
-Headlights projector addition
-Collision fixes so it sits more like it does in UT3 https://drive.google.com/open?id=1DZCJnLy7IRMpJzHNN8loy-05YxL0lbm0 (the only slight issue with this I've noticed is on ONS-Urban where it spawns in the ground and raises up to sit on it's turret, although before the turret was just staying in the ground so maybe this is better anyway?  It doesn't seem to negatively affect use.)

### DIFF
--- a/Classes/UT3Cicada.uc
+++ b/Classes/UT3Cicada.uc
@@ -240,7 +240,25 @@ defaultproperties
     //==========END==========
     VehicleNameString = "UT3 Cicada"
 
-    GroundSpeed=2000
+    GroundSpeed=1600
+    MaxRandForce=2.0
+    RandForceInterval=0.95
+    RollTorqueMax=60 //100
+    RollTorqueStrafeFactor=130 /100
+    RollTorqueTurnFactor=250 //750
+    RollDamping=100.0 //30.0
+    PitchTorqueFactor=50.0 //200.0
+    PitchTorqueMax=20.0 //35.0
+    TurnTorqueFactor=600.0
+    TurnTorqueMax=220.0 //200.0
+    TurnDamping=65.0 //50.0
+    MaxYawRate=1.8 //1.5
+    MaxRiseForce=130.0 //200
+    UpDamping=0.08 //0.05
+    MaxStrafeForce=45.0 //65.0
+    LatDamping=0.08 //0.05
+    MaxThrustForce=80.0 //80.0
+    LongDamping=0.8 //0.3
     CollisionHeight=70.000000
     //DriverWeapons(0)=(WeaponClass=class'UT3CicadaMissileLauncher',WeaponBone=RightRLAttach);
     //DriverWeapons(1)=(WeaponClass=class'UT3CicadaMissileLauncher',WeaponBone=LeftRLAttach);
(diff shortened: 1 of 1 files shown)


Classes/UT3Cicada.uc
```diff
@@ -240,20 +240,60 @@ defaultproperties
     //==========END==========
     VehicleNameString = "UT3 Cicada"
 
-    GroundSpeed=2000
+    GroundSpeed=1600
+    MaxRandForce=2.0
+    RandForceInterval=0.95
+    RollTorqueMax=60 //100
+    RollTorqueStrafeFactor=130 /100
```
GreatEmerald: Clearly a typo, this would mean division...

Classes/UT3Cicada.uc
```diff
@@ -240,20 +240,60 @@ defaultproperties
...
     CollisionHeight=70.000000
     //DriverWeapons(0)=(WeaponClass=class'UT3CicadaMissileLauncher',WeaponBone=RightRLAttach);
     //DriverWeapons(1)=(WeaponClass=class'UT3CicadaMissileLauncher',WeaponBone=LeftRLAttach);
     TargetLockSound=sound'UT3A_Vehicle_Cicada.Sounds.A_Vehicle_Cicada_TargetLock01'
     IdleSound=Sound'UT3A_Vehicle_Cicada.Sounds.A_Vehicle_Cicada_EngineLoop02'
     StartUpSound=Sound'UT3A_Vehicle_Cicada.Sounds.A_Vehicle_Cicada_Start01'
     ShutDownSound=Sound'UT3A_Vehicle_Cicada.Sounds.A_Vehicle_Cicada_Stop01'//8/10
+    ImpactDamageMult = 0.00003 //0.0003
+    DamagedEffectHealthSmokeFactor=0.65 //0.5
+    DamagedEffectHealthFireFactor=0.38 //0.25
```
GreatEmerald: This should be 0.4, see source code: http://greatemerald.xmpcommunity.com/uncodex/UT3-15/Source_utgame/utvehicle.html#6953

Classes/UT3Cicada.uc
```diff
@@ -240,20 +240,60 @@ defaultproperties
...
+    DamagedEffectFireDamagePerSec=0.95 //0.75
```
GreatEmerald: As odd as it may be, it is in fact 2.0 in the code: http://greatemerald.xmpcommunity.com/uncodex/UT3-15/Source_utgame/utvehicle.html#7021

GreatEmerald:
About the binary file: can this be done in code instead?

Classes/UT3Cicada.uc
HellDragon-HK: Actually no, some vehicles do have different damage rates and I had to go up and down to get the exact, within the same 6 damage in each game it has to be right?  I think the SPMA was one of the cases where it needed to be higher than normal, smoke on the other hand was pretty much always the same but fire wasn't not sure why.

Yeah my mistake on the / I'll fix it

Classes/UT3Cicada.uc
HellDragon-HK: So it wasn't my imagination some are losing 2 points of health per second after all, it's really hard to tell from actual game play between the get in and get out time it takes damage so quickly in UT3

HellDragon-HK:
It expands the collision box and changes z translation, I guess z translation does work as prepivot but no idea about collision boxes and the one from the UT3 code is unrecognized on any collision type it had from what I tried

GreatEmerald:
Yea, again, be very careful with Z translation because that changes the centre of mass! I'd prefer if that was done in code via `PrePivot` and `CollisionHeight`/`CollisionRadius`.

Since everything else is fine now, I'll accept this. I'm going to put off pulling in binary packages for now, so if you can find a way to do that through code rather than through editing the file, open a new pull request.

HellDragon-HK:
That's fine, the binary files need to be tested more first, they are being listed so I'm not the only one who has access to the edits

PR #66, opened 2017-12-22
UT3 Goliath Machinegun Turret Update

HellDragon-HK:
Pitch adjusted to be closer to UT3, weapon fire adjusted to be closer to weapon barrel and not as floaty, scaling damage type was needed and can be found here https://drive.google.com/open?id=11uMpAiP9SUgkJWbzIhwQwzvTh87PMDNc , spread and damage changed to UT3 values,  aim error UT3 value added.


GreatEmerald:
Again, include the .uc in your pull request.

GreatEmerald:
You deleted `Classes/UT3GoliathTurret.uc`.

HellDragon-HK:
What how? I don't have permission to do that, unless you mean here in the request in which case I thought you picked the commit you wanted and could ignore others, ok so I need to go back through and undo what I did for trying to add the damage types?

PR #67, opened 2017-12-22
UT3 Hellbender Overall Improvements

HellDragon-HK:
This is everything I've changed with the Hellbender thus far which includes
-Drawscale (0.95 like Alex said it needed to be unless I misunderstood in which case back to the turret size and corona placement issues)
-Turret Size & Collision Update https://drive.google.com/open?id=1UVMWr1TS6UR3CvhCeTNkxDUJLrUm0wFS 
-Sound Pack Update for missing explosion sound https://drive.google.com/open?id=1seZJkSMcgqS5aNgjNgqlj598RCad_DYV 
--Needs Scorpion sound update as well  https://drive.google.com/open?id=1l7cC58tLhf-nAps3Dzk00FttPPchNaDP
-How much the vehicle is thrown around by weapon fire
-Speed
-Handling
-Exit Positions closer to UT3
-Headlights adjustments for smaller size
-Headlight Projector adjustment
-Brakelight adjustment

### DIFF
--- a/Classes/UT3Hellbender.uc
+++ b/Classes/UT3Hellbender.uc
@@ -423,6 +423,7 @@ defaultproperties
     IdleSound=Sound'UT3A_Vehicle_Hellbender.Sounds.A_Vehicle_Hellbender_EngineIdle01'
     StartUpSound=Sound'UT3A_Vehicle_Hellbender.Sounds.A_Vehicle_Hellbender_EngineStart01'
     ShutDownSound=Sound'UT3A_Vehicle_Hellbender.Sounds.A_Vehicle_Hellbender_EngineStop01'
+    ImpactDamageMult = 0.00005
     DamagedEffectHealthSmokeFactor=0.65 //0.5
     DamagedEffectHealthFireFactor=0.40 //0.25
     DamagedEffectFireDamagePerSec=2.0 //0.75
@@ -456,20 +457,18 @@ defaultproperties
     WheelInertia=0.01
     ChassisTorqueScale=0.82 //0.7
     WheelSuspensionOffset=5.0 //this one little value causes the Hellbender to sit lower like it should, unbeliveble we went crazy trying to solve it and all it needed was this
+
     //Drawscale=1.2 UT3 Draw template, collision needs to be updated with ukx
-    //DrawScale=0.95
+
 
     CollisionRadius=219
     
-    //ExitPositions(0)=(X=0,Y=-165,Z=50)
-    //ExitPositions(1)=(X=0,Y=165,Z=50)
-    //ExitPositions(2)=(X=0,Y=-165,Z=-50)
-    //ExitPositions(3)=(X=0,Y=165,Z=-50)
     
     ExitPositions(0)=(X=-10,Y=-160,Z=50)  //Left
     ExitPositions(1)=(X=-10,Y=160,Z=50)   //Right
     ExitPositions(2)=(X=-10,Y=-160,Z=-50) //Left Below
     ExitPositions(3)=(X=-10,Y=160,Z=-50)  //Right Below
+    ExitPositions(4)=(X=10,Y=-5,Z=130)    //Roof
     
     FPCamPos=(X=0,Y=28,Z=135)
     
--- a/Classes/UT3Hellbender.uc
+++ b/Classes/UT3Hellbender.uc
@@ -58,7 +58,8 @@ function AltFire(optional float F) //This is to remove the horn each time you fi
 
 function TakeDamage(int Damage, Pawn instigatedBy, Vector Hitlocation, Vector Momentum, class<DamageType> DamageType)
 {                            //Make sure you don't hurt yourself with a combo
-    if (InstigatedBy != self)
+    if (InstigatedBy == self && ClassIsChildOf(DamageType, class'DamTypeSkyMine'))
+    return;
     Super.TakeDamage(Damage, instigatedBy, Hitlocation, Momentum, damageType);
 }
 
--- a/Classes/UT3Hellbender.uc
+++ b/Classes/UT3Hellbender.uc
@@ -436,8 +436,8 @@ defaultproperties
     ExplosionSounds(2) = Sound'UT3A_Vehicle_Hellbender.Sounds.A_Vehicle_Hellbender_Explode01';
     ExplosionSounds(3) = Sound'UT3A_Vehicle_Hellbender.Sounds.A_Vehicle_Hellbender_Explode01';
     ExplosionSounds(4) = Sound'UT3A_Vehicle_Hellbender.Sounds.A_Vehicle_Hellbender_Explode01';
-    EntryRadius=300.000000
-    TPCamWorldOffset=(Z=200.000000)
+    EntryPosition=(X=0,Y=0,Z=0)
+    EntryRadius=180.0  //300.000000
     MomentumMult=0.400000 //1.0  //HDm to GE: 0.4 feels right but Rocket and AVRiL force are reversed with each other
     MomentumMult=1.000000
     bDrawDriverInTP=False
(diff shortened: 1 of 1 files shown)


GreatEmerald:
I guess the suspension offset change should be here too. Also, I need to look at the DrawScale (or someone else can test this).

HellDragon-HK:
Suspension fix is already here at line 457

Classes/UT3Hellbender.uc
```diff
@@ -1,6 +1,7 @@
 /*
  * Copyright © 2008, 2014 GreatEmerald
  * Copyright © 2008-2009 Wormbo
+ * Copyright © 2017 HellDragon (still too strong a word but if you want it, here you go)
```
GreatEmerald: Again, no comments allowed in the copyright line.
While you're at it, you can also add `, 2017` to my line.

Classes/UT3Hellbender.uc
```diff
@@ -57,7 +58,8 @@ function AltFire(optional float F) //This is to remove the horn each time you fi
 
 function TakeDamage(int Damage, Pawn instigatedBy, Vector Hitlocation, Vector Momentum, class<DamageType> DamageType)
 {                            //Make sure you don't hurt yourself with a combo
-    if (InstigatedBy != self)
+    if (InstigatedBy == self && ClassIsChildOf(DamageType, class'DamTypeSkyMine'))
+    return;
```
GreatEmerald: Put four more spaces here.

Classes/UT3Hellbender.uc
```diff
@@ -415,33 +418,84 @@ defaultproperties
...
     bDrawDriverInTP=False
     DriverDamageMult=0.000000
     VehiclePositionString="in a Hellbender"
     VehicleNameString="UT3 Hellbender"
     HornSounds(0)=Sound'UT3A_Vehicle_Hellbender.Sounds.A_Vehicle_Hellbender_Horn01'
-    GroundSpeed=700.000000
+    GroundSpeed=800.000000 //700
     SoundVolume=255
 
-    DrawScale=1.0
```
GreatEmerald: By default the `DrawScale` is not 1.0. Yet if you scaled it up in the UKX to the right size, then `DrawScale` should be set to 1.0, because that means "no further changes to scale needed in the code".

Classes/UT3Hellbender.uc
```diff
@@ -415,33 +418,84 @@ defaultproperties
...
+    TransRatio=0.15 //0.11
+    EngineBrakeFactor=0.0002 //0.0001 def
+    MaxBrakeTorque=20.5 //20.0
+    EngineInertia=0.01
+    WheelInertia=0.01
+    ChassisTorqueScale=0.82 //0.7
+    WheelSuspensionOffset=5.0 //this one little value causes the Hellbender to sit lower like it should, unbeliveble we went crazy trying to solve it and all it needed was this
```
GreatEmerald: I suggest not making lines longer than ~100 characters, otherwise it looks cut off

Classes/UT3Hellbender.uc
HellDragon-HK: Sure, do you want me to try and catch that on any others too along with GP?

Classes/UT3Hellbender.uc
HellDragon-HK: It might be a good idea to set that as default yes because some of the stock classes have a drawscale setting we don't want interferring right?

Classes/UT3Hellbender.uc
HellDragon-HK: I can remove the comment, just upon finding that one little key value that allowed fixing every ground vehicle was one of those kind of moments know what I mean

Classes/UT3Hellbender.uc
GreatEmerald: No. The year indicates when the person contributed. I contributed the way to add the function here, hence the year should be updated, but nowhere else.

Classes/UT3Hellbender.uc
GreatEmerald: Indeed. You could double-check all these to make sure they're all 1.0, but just setting it to 1.0 makes it certain.

Classes/UT3Hellbender.uc
GreatEmerald: Sure, it's fun to document things like that, just perhaps not use that much space for it.

Classes/UT3Hellbender.uc
GreatEmerald: Also, don't update it like that, it needs to say `2008, 2014, 2017`!

HellDragon-HK:
What other changes here are still being needed?  I set scale to a forced 1.0, it's just been moved up to where I usually have it which on this one is line 330.

GreatEmerald:
My copyright line; you should not remove 2014 from it.

GreatEmerald:
OK, you're missing the changes to defaultproperties. If you check the logs you'll see a whole lot of `Accessed None`s with regards to the Hellbender weapons, that's because the array of weapons is longer due to the vanilla Hellbender having more of them. That's probably what's causing this weird behaviour with the number buttons as well. I'm not sure what works, though, we need to experiment a bit.

First of all, remove `PassengerWeapons(1)=()`. Then try adding `PassengerWeapons=()` above `PassengerWeapons(0)=(WeaponPawnClass=Class'UT3HellbenderRearGunPawn',WeaponBone="MainTurretYaw")`. If that doesn't fix the accessed nones and the strange button behaviour, then try replacing `PassengerWeapons(0)=(WeaponPawnClass=Class'UT3HellbenderRearGunPawn',WeaponBone="MainTurretYaw")` with `PassengerWeapons=(WeaponPawnClass=Class'UT3HellbenderRearGunPawn',WeaponBone="MainTurretYaw")` or `PassengerWeapons=(0=(WeaponPawnClass=Class'UT3HellbenderRearGunPawn',WeaponBone="MainTurretYaw"))`.

HellDragon-HK:
The first option there is what I already had and just missed it in the request, the 2nd and 3rd I can't get to compile

redundant data:     PassengerWeapons=(0=(WeaponPawnClass=Class'UT3HellbenderRearGunPawn',WeaponBone="MainTurretYaw"))


    MovementAnims(0) = "Idle"

    DriverWeapons(0)=(WeaponClass=Class'UT3HellbenderSideGun',WeaponBone="SecondaryTurretYaw")
    PassengerWeapons=()
    PassengerWeapons=(0=(WeaponPawnClass=Class'UT3HellbenderRearGunPawn',WeaponBone="MainTurretYaw"))
    
    FlagBone = Hood;

GreatEmerald:
Ah yes, it says it's redundant because of the `PassengerWeapons=()` line, remove that one while testing this.

HellDragon-HK:
Uh the Hellbender has a turret vacancy

GreatEmerald:
Eh?

HellDragon-HK:
I mean it has no rear turret now

GreatEmerald:
Huh, that's weird. So what happens with:
```
PassengerWeapons=(WeaponPawnClass=Class'UT3HellbenderRearGunPawn',WeaponBone="MainTurretYaw")
```
And with:
```
PassengerWeapons=((WeaponPawnClass=Class'UT3HellbenderRearGunPawn',WeaponBone="MainTurretYaw"))
```

HellDragon-HK:
PassengerWeapons=(WeaponPawnClass=Class'UT3HellbenderRearGunPawn',WeaponBone="MainTurretYaw")

Still vacant

PassengerWeapons=((WeaponPawnClass=Class'UT3HellbenderRearGunPawn',WeaponBone="MainTurretYaw"))

This fixed it and no number 3 key either now nor respawn on back turret

GreatEmerald:
Excellent! So *that* is a great example of how to get rid of a turret! Turns out that you need one pair of `()` to show that this is an array, and the other pair of `()` inside the first one to show that this is the one and only element inside it.

HellDragon-HK:
Added in the 'Great example of how to get rid of a turret' (without actually saying it is, do you want that added as well for future reference on what was done in case it's needed later?)

Added in the team corona lines

GreatEmerald:
Yea, move that comment over there. And also remove all the lines that you now have commented out, they are no longer needed.

HellDragon-HK:
Changed, is that how you wanted it?

Classes/UT3Hellbender.uc
```diff
@@ -415,33 +302,80 @@ defaultproperties
...
+    ImpactDamageSounds(2) = Sound'UT3A_Vehicle_Scorpion.Sounds.A_Vehicle_Scorpion_Collide03';
+    ImpactDamageSounds(3) = Sound'UT3A_Vehicle_Scorpion.Sounds.A_Vehicle_Scorpion_Collide04';
+    ImpactDamageSounds(4) = Sound'UT3A_Vehicle_Scorpion.Sounds.A_Vehicle_Scorpion_Collide03';
+    ImpactDamageSounds(5) = Sound'UT3A_Vehicle_Scorpion.Sounds.A_Vehicle_Scorpion_Collide04';
+    ImpactDamageSounds(6) = Sound'UT3A_Vehicle_Scorpion.Sounds.A_Vehicle_Scorpion_Collide03';
+    ExplosionSounds(0) = Sound'UT3A_Vehicle_Hellbender.Sounds.A_Vehicle_Hellbender_Explode01';
+    ExplosionSounds(1) = Sound'UT3A_Vehicle_Hellbender.Sounds.A_Vehicle_Hellbender_Explode01';
+    ExplosionSounds(2) = Sound'UT3A_Vehicle_Hellbender.Sounds.A_Vehicle_Hellbender_Explode01';
+    ExplosionSounds(3) = Sound'UT3A_Vehicle_Hellbender.Sounds.A_Vehicle_Hellbender_Explode01';
+    ExplosionSounds(4) = Sound'UT3A_Vehicle_Hellbender.Sounds.A_Vehicle_Hellbender_Explode01';
```
GreatEmerald: All of these should also be folded into single arrays. Put `ExplosionSounds=()` and `ImpactDamageSounds=()` above and remove the duplicate entries.

GreatEmerald:
Oh, also, I hope that you're not seeing `Accessed None`s in the log with regards to the Hellbender any more now?

HellDragon-HK:
What do you mean by status of binaries?   Every binary file I've worked with to date is on G-Drive currently I'm fairly sure.

I'm not, that was something I noticed right away

GreatEmerald:
Uh, that's not quite how you are supposed to clear the arrays.

So all the binaries you have in `Binaries.csv` already refer to the scale at which the vehicles need to be for the headlight offsets etc. to be correct?

HellDragon-HK:
How do I need to do it then?

Should be yes

GreatEmerald:
I said `ExplosionSounds=()` and `ImpactDamageSounds=()`.

Yea, the manta's headlight in your screenshot looked weird, but that's for another issue I suppose.

GreatEmerald:
OK, thanks.

PR #68, opened 2017-12-22
UT3 Hellfire SPMA Overall Improvement

HellDragon-HK:
This is everything I've changed with the Hellfire SPMA which includes..
-Smoke & Fire Damage Phases in-line with UT3
-Sound Pack Update for SPMA https://drive.google.com/open?id=17UqS7l3-Jx-lbN34S0QrLVglb6WH2Gxh  (With reduced turret rotate volume so it's less annoying, you'll probably want this @Unre-Alex)
--Can't remember if Collide was in the Goliath uax or not, if not here is the updated Goliath which has missing sounds added anyway https://drive.google.com/open?id=1uFWoDVvJpaGgDzxAYA8KkGT6Ija4CRKI
-Speed
-Handling
-How much the vehicle is thrown by weapon fire
-Removed Daredevil as the Hellfire SPMA is more of a tank and shouldn't qualify for it
-Exit Positions closer to UT3 (Alex said he had issues, I haven't though and all I did was move it slightly forward to be where the door is and closer to ground which both of these are in-line with UT3 from my testing)
-Headlights & Projector disabled due to floating

### DIFF
--- a/Classes/UT3HellfireSPMA.uc
+++ b/Classes/UT3HellfireSPMA.uc
@@ -924,17 +924,79 @@ defaultproperties
     IdleSound      = Sound'UT3SPMA.SPMAEngineIdle'
     StartUpSound   = Sound'UT3SPMA.SPMAEngineStart'
     ShutDownSound  = Sound'UT3SPMA.SPMAEngineStop'
-
+    DamagedEffectHealthSmokeFactor=0.65 //0.5
+    DamagedEffectHealthFireFactor=0.39 //0.25
+    DamagedEffectFireDamagePerSec=0.95 //0.75
+    ImpactDamageSounds(0) = Sound'UT3A_Vehicle_Goliath.Sounds.A_Vehicle_Goliath_Collide01';
+    ImpactDamageSounds(1) = Sound'UT3A_Vehicle_Goliath.Sounds.A_Vehicle_Goliath_Collide01';
+    ImpactDamageSounds(2) = Sound'UT3A_Vehicle_Goliath.Sounds.A_Vehicle_Goliath_Collide01';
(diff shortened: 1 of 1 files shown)


Classes/UT3HellfireSPMA.uc
```diff
@@ -924,17 +924,79 @@ defaultproperties
     IdleSound      = Sound'UT3SPMA.SPMAEngineIdle'
     StartUpSound   = Sound'UT3SPMA.SPMAEngineStart'
     ShutDownSound  = Sound'UT3SPMA.SPMAEngineStop'
-
+    DamagedEffectHealthSmokeFactor=0.65 //0.5
+    DamagedEffectHealthFireFactor=0.39 //0.25
+    DamagedEffectFireDamagePerSec=0.95 //0.75
```
GreatEmerald: Same point here with 0.4 and 2.0

Classes/UT3HellfireSPMA.uc
```diff
@@ -924,17 +924,79 @@ defaultproperties
...
+    HandbrakeThresh=9000000
+    MinBrakeFriction=6.5  //4.0
+    EngineBrakeRPMScale=0.008
+    TransRatio=0.18
+    ChangeUpPoint=105000
+    
+    MomentumMult=0.1 //2.0
+    bDoStuntInfo=False //true
+    
+        Begin Object Class=KarmaParamsRBFull Name=KParams0
```
GreatEmerald: Same point here with aligning the begin with the end

HellDragon-HK:
There you go, how is it that the Begin part was moved, it's right in the uc and I copy and paste rather than type it in to avoid mistakes

GreatEmerald:
GitHub editor is weird with pasting sometimes.

Thanks, this looks OK now.

PR #69, opened 2017-12-22
UT3 Leviathan Overall Improvements

HellDragon-HK:
Not much here, it already handles like it does in UT3 I think, what has been changed includes...
-Speed
-Impact Damage
-Smoke & On Fire points (This is NOT in-line with UT3 as I feel it starts way too early, it's already on fire with at least 5100 health left in UT3, in UT2004 it's more toward 3000-2000 with these values which seems more reasonable)
-Exit Points
-How much it's thrown around
-Camera Views (There is an unlisted camera view somewhere that I can't edit yet which is getting out while moving drops you in the middle of the Levi just behind the front wheels)
-Headlights
-Headlight Projector
-Collision & Turret Size Update https://drive.google.com/open?id=1_2foPegkhCEcJ2hvyxMGEF7kD0fNqJSa
 (Driver Turret was too small and collision changes help Levi not bottom out)

### DIFF
--- a/Classes/UT3Leviathan.uc
+++ b/Classes/UT3Leviathan.uc
@@ -404,6 +404,7 @@ defaultproperties
     PassengerWeapons(3) = (WeaponPawnClass=class'UT3LeviathanTurretShock',WeaponBone="RT_Rear_TurretYaw")
 
     CollisionHeight=100.0
+    GroundSpeed=600
     LSDFactor=1.000000
     ChassisTorqueScale=0.200000
     MaxSteerAngleCurve=(Points=((OutVal=30.000000),(InVal=1500.000000,OutVal=20.000000)))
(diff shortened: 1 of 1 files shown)


GreatEmerald:
Looks good, but do use the UT3 values for fire. It was set like that on purpose for balancing reasons, looks like.

HellDragon-HK:
Ah yeah I had a feeling you were going to say that, my only complaints are it doesn't make sense with that much health and it's deceptive for an attacker because anything on fire should be an easy target but that probably won't be

HellDragon-HK:
Well there it is, damage phases changed.....but hear me out, what I think is it was set early for testing the effects and never changed, there are several things I've noticed about the Levi that support this theory, it has floating smoke exhaust that should have had a Z value but I'm guessing only has a quick guess X value, the driver turret looks like it was meant to have a light (or a scope look maybe) but doesn't, and the circles on the back of it resemble the Hellbender's reverse brake lights yet they never light up, it also has floating and kind of oddly placed smoke for it's damage phase, to me it seems like it was made game play ready but the visual special effects were still being worked on and never completed

GreatEmerald:
Looks good. The damage phases could be adjusted sometime later. I still think it's a balancing thing, to make it respawn quicker at the base if someone (i.e. bot) leaves it empty.

HellDragon-HK:
Trust me it respawns well before it can take much damage so I don't think so

PR #71, opened 2017-12-22
Update UT3LeviathanDriverWeapon.uc

HellDragon-HK:
Slowed turret rotation closer to UT3 and re-aligned weapon fire for turret size increase


GreatEmerald:
This requires an updated file for the meshes, doesn't it? Can you scale it up in code rather than doing it in the mesh?

HellDragon-HK:
With the UT3 Drawscale these re-alignments will need to be re re-aligned I think first

GreatEmerald:
So you're saying I shouldn't merge this yet, then?

HellDragon-HK:
Well it works for what the current archive is so I'm not really sure, on my end it probably needs to be redone because the Levi is now radically different than the archive....so up to you I guess?

GreatEmerald:
There's no point to changing it now if it will need to be changed again in the future, then. Update this request once you have the numbers for the UT3 DrawScale.

HellDragon-HK:
Redone

PR #72, opened 2017-12-22
Update UT3LeviathanPrimaryWeapon.uc

HellDragon-HK:
Rotation speed closer to UT3 and new DmgType to include different values UT3 has from UT2004 which are VehicleDamageScaling, KDamageImpulse, and KDeathUpKick https://drive.google.com/open?id=1uUtrXBRj9ZtaU7u8BInslmPk4kPSZ8OD   I also couldn't add KImpulseRadius at all as it won't compile

(I couldn't find the damage for the main cannon in UT3, the only thing I found is this but it can't be right as a Goliath in UT3 at full 900 health is destroyed in a single shot so it's doing more than this code says it does)
	ExplosionDamage=150.0
	ExplosionRadius=1000.0
	ExplosionMomentum=125000

And here is what I think we currently have from default UT2004
    DamageMin=1000
    DamageMax=1000
    DamageRadius=2200
    Momentum=2000000.0


GreatEmerald:
Include that .uc in the pull request.

The explosion damage works the same way as the Redeemer: it deals damage several times in rapid succession to simulate a shockwave. See http://greatemerald.xmpcommunity.com/uncodex/UT3-15/Source_utgamecontent/utemit_leviathanexplosion.html#101

HellDragon-HK:
Ok I see that and I also see this
	Damage=250.0
	DamageRadius=2000.0
	MomentumTransfer=250000

Does anything need to be done, I've replicated destroying a power core in a single shot....at long distacne even on ONS-ColdFusion (which is an upcoming video and you can see the shot for yourself, I couldn't even see the power core from where I fired)

GreatEmerald:
The files are completely messed up here now. It shows that you renamed a file and replaced all its contents, which is absolutely not what you want to do. For that matter, it might be faster if you start over with this pull request...

But yes, to make the damage equal, you'd need to override the function `Explosion()` to replace this line: http://greatemerald.xmpcommunity.com/uncodex/UT2004/Source_onslaughtfull/onsmascannon.html#155 With:
```UnrealScript
damageScale = 1
```
Right now it's made in a way that it does damage relative to how close you are in relation to the centre of the explosion. What it should be doing is causing constant damage always, but the lower amount (250 instead of 1000).

PR #73, opened 2017-12-22
Update UT3LeviathanShield.uc

HellDragon-HK:
Shield visual and sound change to match UT3, ambientsound is not working though

### DIFF
--- a/Classes/UT3LeviathanShield.uc
+++ b/Classes/UT3LeviathanShield.uc
@@ -6,8 +6,9 @@ Last change: $Id$
 Copyright (c) 2007, Wormbo
 ******************************************************************************/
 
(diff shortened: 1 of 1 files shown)


GreatEmerald:
The ambient should should work. Maybe try setting `AmbientSoundVolume=255`. Usually the ambient sounds have rather limited volume.

Does this require updates to binary files?

HellDragon-HK:
I don't think so as I'm pretty sure the shield sounds were already in the Paladin uax so it's just a path change

HellDragon-HK:
UT3Vehicles.UT3LeviathanShield: Unknown property in defaults:     AmbientSoundVolume=255

GreatEmerald:
Oh, it's just `SoundVolume=255`.

PR #74, opened 2017-12-22
Update UT3LeviathanShieldEffectBlue.uc

HellDragon-HK:
Visual and size change to match UT3 a little more
NOTE: Don't forget Red and HitEffect variants as well

### DIFF
--- a/Classes/UT3LeviathanShieldEffectBlue.uc
+++ b/Classes/UT3LeviathanShieldEffectBlue.uc
@@ -9,7 +9,7 @@ class UT3LeviathanShieldEffectBlue extends Emitter;
 DefaultProperties
 {
     Begin Object Class=MeshEmitter Name=MeshEmitter18
-        StaticMesh=StaticMesh'AW-2k4XP.Weapons.ShockShield2'
+        StaticMesh=StaticMesh'UT3PaladinSM.PaladinShield' //'AW-2k4XP.Weapons.ShockShield2'
         UseParticleColor=True
         UseColorScale=True
         AutomaticInitialSpawning=False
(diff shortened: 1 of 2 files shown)


GreatEmerald:
Looks OK, but please also include the red variant to this pull request as well.

HellDragon-HK:
Last time you deleted it as a duplicate so I didn't, so I am supposed to do both then?

GreatEmerald:
Either works. It's more convenient for me if you close it yourself.

Anyway, the same question here: are you sure X here should be 0.7 and not 1.0?

HellDragon-HK:
Yes, the shield needs to go about halfway across the Levi, and it does it as far as I can tell with this setting (with the visuals being different it's hard to tell), using the UT3 drawscale I rather doubt it's enough and 1.0 may not do it either.........wait I'm confusing X & Y aren't I....uh ok if this is how far out it is from the turret's front then no I can't really tell with the visuals but it looks close, it probably needs, actually if you accept the UT3 drawscale will need to be looked at again

GreatEmerald:
Yea, so testing it with 1.0 on all three axes would be nice.

HellDragon-HK:
I could use some help confirming the size, I think the values I've got are close though
I've been using this image as well as 2 others from different angles for reference to try and scale it
screenshot00001

GreatEmerald:
Looking at the code:
http://greatemerald.xmpcommunity.com/uncodex/UT3-15/Source_utgamecontent/utpaladinshield.html
http://greatemerald.xmpcommunity.com/uncodex/UT3-15/Source_utgamecontent/utleviathanshield.html
...those are different meshes altogether.

HellDragon-HK:
I guess this explains the scale difference then?
There doesn't appear to be a Leviathan usx file yet so what now?

PR #75, opened 2017-12-22
Update UT3LeviathanShockBall.uc

HellDragon-HK:
Uh why was the base damage at 120 when the code says it's only 45, vehicles may take less but players don't I think maybe?


GreatEmerald:
Looks good, thanks.
I think this code was written by Wormbo. He likes aligning all defaultproperties to the equals sign (which I don't consider a good idea, because you have a problem if you ever use a longer name for a defaultproperty).

PR #76, opened 2017-12-22
Update UT3DmgType_LeviathanShockBall.uc

HellDragon-HK:
Damage Scaling set to match UT3, odd .25 is what is in UT3 code but doesn't come close to the same damage, .68 gives the same damage as in UT3

### DIFF
--- a/Classes/UT3DmgType_LeviathanShockBall.uc
+++ b/Classes/UT3DmgType_LeviathanShockBall.uc
@@ -46,6 +46,6 @@ defaultproperties
     FemaleSuicide="%o recieved too large of a shock."
 
     VehicleMomentumScaling=0.25
-    VehicleDamageScaling=0.25
+    VehicleDamageScaling=0.68 //0.25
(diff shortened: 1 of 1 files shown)


GreatEmerald:
Hm, interesting. The reason it seems that way is actually quite different: http://greatemerald.xmpcommunity.com/uncodex/UT3-15/Source_utgamecontent/utproj_leviathanshockball.html#60

What it says is that every time it explodes it causes a combo, which in turn uses the 120 damage value. So the good way to implement this would be to change the shock ball to always do combos in UT2004 as well.

HellDragon-HK:
It is always doing combos in 2004 though that I can see, it's a combo explosion visual as soon as it hits something

GreatEmerald:
Oh, I see where the problem is. Fixed in commit https://github.com/GreatEmerald/UT3Vehicles/commit/fc9007fed10b953f14027cfe01ce28a3e2a1da2e.

PR #77, opened 2017-12-22
UT3 Leviathan Turret Sound Fix & Driver Position Improvement

HellDragon-HK:
Fixed missing sound when entering and leaving a turret as well as improved the driver position a little


GreatEmerald:
Does this require changes to binaries?

HellDragon-HK:
I don't have a backup of the Levi uax so I don't think so as it already has everything it needs and those just were not used, I could be mistaken but I also don't see a uax for this on my G-Drive so I'm going to say no it doesn't

GreatEmerald:
OK, thanks.

PR #78, opened 2017-12-22
UT3 Leviathan Beam Turret Exit Positions and Cam Improved

HellDragon-HK:
Exit Positions were the exact same as the Levi itself, UT3 has different points and this is closer to it along with a bit of camera tweaking


GreatEmerald:
Looks good, thanks.

PR #79, opened 2017-12-22
UT3 Leviathan Rocket Turret Exit Points and Cam Improved

HellDragon-HK:
Exit points set closer to UT3 and a little bit of camera tweaking (NOTE: Not a duplicate, each turret needed it's own tweaks in both exit points and camera)


GreatEmerald:
OK, thanks.

PR #81, opened 2017-12-22
UT3 Leviathan 'Stinger' Turret Exits & Cam Improved

HellDragon-HK:
Exit points closer to UT3 & some camera tweaking (NOT a duplicate)


GreatEmerald:
OK, thanks.

It's interesting that each turret has its own exit points, but that does make sense in the context of the Leviathan.

HellDragon-HK:
It's also the case because from what I can tell UT3 doesn't use exit points it uses exit radius, although it does seem to drop you in the same spot so maybe it does and it's hidden somewhere or I just overlooked it

PR #84, opened 2017-12-22
Update UT3LeviathanTurretWeaponMinigun.uc

HellDragon-HK:
Scaling needed to be adjusted
New DmgType Here
https://drive.google.com/open?id=1MtROHsIV7qhAs6HazsrD5Qcqmp9IzGzT


GreatEmerald:
You don't seem to be doing any scaling here, though?
As for the sound, can that be confirmed? Indeed mixups may have happened.
And include the .uc in the pull request too.

HellDragon-HK:
the scaling is in the damage type and hold on I'll go double check the sound

HellDragon-HK:
Yeah confirmed it's alt fire but more rapid in frequency than the normal handheld Stinger

GreatEmerald:
Again, you destroyed the entirety of the `Classes/UT3LeviathanTurretWeaponMinigun.uc` file!

PR #85, opened 2017-12-22
UT3 Paladin Shield Size Closer to UT3

HellDragon-HK:
This resizes the Paladin's shield to be closer to UT3
NOTE: Don't forget Red


GreatEmerald:
Are you sure about the scaling? I don't expect it to be different on different axes.
And include the red in the same pull request.

HellDragon-HK:
You know with the UT3 Drawscale now that probably is the case this will need to be reworked but Alex was pretty sure of it yes

GreatEmerald:
OK, not merging yet then. Would be nice if someone could confirm the aspect ratio here.

Unre-Alex:
@HellDragon-HK I didn't say I was sure, I said the Z axis needs to change. If that's for the drawscale 1.0 than the Z should be like Y, 1.5. For the drawscale 1.35 just add 0.35 to all the axes.

HellDragon-HK:
Ok sorry, hm just add .35 to all, does that actually work?

Unre-Alex:
@HellDragon-HK I thought you can change the scale in the usx file, but you can't. In that case the scale for the 1.35 would have to be something like X 4.0 Y & Z 2.0

Did you notice that coronas disappear behind the shield?

HellDragon-HK:
That's larger than what I got for the Levi shield which is 1.4 Y & 1.7 X though that can't be right can it?

No but which coronas, I know the shields in UT3 have 2 coronas of their own that appear but if you mean in UT2004 headlight coronas no I didn't

Unre-Alex:
I haven't looked into the Levi shield yet.

In UT2004, all kinds of coronas disappear.

HellDragon-HK:
You said you were having the same issue with projector though, I haven't had that even on a LAN game with my nephew

Unre-Alex:
It's not that, just activate the shield and look through it where there is any corona.

HellDragon-HK:
You're right it disables as shield goes over, it fades out so more than likely it's not a bug and the way the game handles it

HellDragon-HK:
@Unre-Alex Well Alex you almost got the start size right with the 4.0 guess lol uh kind of

screenshot00000
shot00612

Unre-Alex:
4.0 for X not Y

HellDragon-HK:
EDIT - My mistake, misread Y as 4.0 when you said it was & Z of 2.0 which gives us this
shot00613
screenshot00002pal

Unre-Alex:
The cannon should be 1.35 too, right?

HellDragon-HK:
It should in theory but it was way too small and I think I ended up at 1.9 for it but you can double check

Unre-Alex:
Right 1.35 is tiny, 1.7 looks good. Where did you find the drawscale for the vehicles? If in the code than there's got to be a drwascale for the cannon.

HellDragon-HK:
The weapons are part of the vehicle code in UT3 so yes I agree there should be a drawscale for them but no luck finding 2 scales, just the 1

btw I can't tell on how far out this is (X value) what do you think?  Can't get the same camera angle no matter how hard I try sorry

screenshot00011
shot00619

Unre-Alex:
Did you use find in files search option in n++?

It's hard to tell, you can record a demo in UT2004 and take a screenshot.

HellDragon-HK:
Wait a second is this doing it?

   DrawScale=0.8 in OnslaughtBP.ONSShockTankCannon

HellDragon-HK:
'Did you use find in files search option in n++?'

What? oh ok I don't have direct access to UT3 uc files, I'm using http://greatemerald.xmpcommunity.com/uncodex/UT3-15/utgamecontent/utvehicle_paladin.html

Unre-Alex:
Add a drawscale to UT3PaladinShieldEffect classes.

HellDragon-HK:
the 0.8 I meant is why the 1.35 making the turret smaller perhaps

Unre-Alex:
Do ONSShockTankShieldEffectBlue and ONSShockTankShieldEffectRed have a drawscale?

HellDragon-HK:
Ah I'm mistaken again DrawScale3D = (X=1.0,Y=1.0,Z=1.0) is in the Levi shield and not the Paladin shield

HellDragon-HK:
Yeah the 0.8 was making the turret smaller, add the Drawscale = 1.0 to the UT3PaladinCannon class then scale it back down to 1.35 in the ukx and it's right

HellDragon-HK:
@Unre-Alex Managed to finally get a screenshot closer to what I have of UT3, and it looks like the X value needs to be decreased to me but what do you think?  Edited to add it above so it's easier to compare

HellDragon-HK:
Well I found out not to try adding any kind of drawscale to the shield, doing so and the shield appears to be there but doesn't stop anything from getting through

Unre-Alex:
I'd stay with 4.0, look how far the shield is in front of the Paladin.

Yes, needs to be moved up and also to the left.

HellDragon-HK:
Ok then so @GreatEmerald how do we move the shield, I don't think I've seen an offset of any kind for it.

GreatEmerald:
Oh, so X actually affected how "far away" the shield appears to be? Yea, obviously, that's not right because what it would actually do is warp the shield so that it becomes "fatter", but since it's attached, it would look like it's further away. The actual way to achieve that would indeed be by changing the offset at which the shield effect is spawned. That is also needed to make sure the collision matches.

The way the collision works is in fact that there is an invisible shield attached to the cannon at all times. Its collision is disabled normally, and enabled (plus the visual effect is spawned) when you activate the shield. The collision is probably read from the static mesh. The scale of the invisible attached mesh (`DrawScale3D`: http://greatemerald.xmpcommunity.com/uncodex/UT2004/Source_onslaughtbp/onsshocktankshield.html#85) must match the scale of the effect (`StartSizeRange`: http://greatemerald.xmpcommunity.com/uncodex/UT2004/Source_onslaughtbp/onsshocktankshieldeffectred.html#21).

And yes, the coronas are invisible through the shield because the shield collides. The game treats it as a solid object and it thinks that lights should not shine through solid objects.

What I don't get is why they use an emitter for it to begin with. The mesh is already there. Why go through the trouble of making a mesh emitter and attaching it, if one could just make the texture on the shield visible or invisible?

HellDragon-HK:
Still not seeing an offset anywhere

GreatEmerald:
For the hidden mesh, it's mostly likely `PrePivot`. It might work for the particles as well. Else the mesh would need its translation adjusted.

HellDragon-HK:
PrePivot seems to have no affect in UT3PaladinShield.uc

HellDragon-HK:
Seems like those changes might work but don't merge until @Unre-Alex gives his opinion

GreatEmerald:
Indeed. However, your `StartSizeRange`s should be `(X=(Min=2.200000,Max=2.200000),Y=(Min=2.2000000,Max=2.2000000),Z=(Min=2.200000,Max=2.200000))` as far as I can tell.

HellDragon-HK:
I did set it to that....did I miss setting it here in the request?

GreatEmerald:
Yeap, now it looks right. But yea, the collision needs to be confirmed, I'm not sure if `PrePivot` affects it. If it doesn't, the binary will need to be changed.

HellDragon-HK:
It works, I already made sure to test before even changing the request but if you need @Unre-Alex can you confirm it too?

Unre-Alex:
Collision works, `PrePivot` needs to change to `X=-100,Y=30.0,Z=-20`

HellDragon-HK:
I know it's close in the front but -100 seemed pretty far away when aimed toward back or side, are you sure?

Unre-Alex:
I was considering more than 100, so...

HellDragon-HK:
Alright then so -100 or even more?

Unre-Alex:
I'm good with 100, it's up to you now.

Unre-Alex:
But it's not fully to the ground in UT3, is it?

HellDragon-HK:
In Islander without moving up or down it's a little above the barricades to the side of the Paladin so I guess it is up in the air a little

GreatEmerald:
Hmm, UT3 says: http://greatemerald.xmpcommunity.com/uncodex/UT3-15/Source_utgamecontent/utpaladinshield.html#57 `(X=400,Z=140)`
I'm guessing that these numbers are assuming that the shield is in the very centre of the coordinate system, and in the binaries the shield had translation applied to it. If you reverse the translation, the numbers should check out.

HellDragon-HK:
Well Up is Down in UT2004 so that should be -400 (positive brings it closer to Paladin) and -140 which can't be right because it looks like this...
shot00634
shot00635

GreatEmerald:
Yes, but how does it look in the editor?

HellDragon-HK:
There doesn't seem to be a way to use translation for a static mesh but it looks as centered as it does in UT3 editor.

GreatEmerald:
OK, then I don't get why there's such a difference. There's also `(X=-50,Z=-20)` in the code for the particles, which seems more like it, but I'd imagine this is more for the projector beams rather than the shield itself...

HellDragon-HK:
Ok rebooted in new request

PR #86, opened 2017-12-22
UT3 Raptor Overall Improvements

HellDragon-HK:
This is everything I've changed on the Raptor so far
-Speed
-Handling
-How much it's thrown around by weapon fire
-Exit Points
-Headlights adjustment
-Projector addition
-Collision https://drive.google.com/open?id=1GAGc8gKLVdO66snC_5j2LVyGBn9QxSTj


GreatEmerald:
Same comments as for the other pulls: fire numbers don't match UT3, and you should align `Begin`s with `End`s.

GreatEmerald:
OK, thanks.

PR #87, opened 2017-12-22
UT3 Scorpion Overall Improvements

HellDragon-HK:
This is everything I've changed with the Scorpion for now, which includes...
-Fixed boost sound on 32 bit
-Handling
-Sound Pack Update https://drive.google.com/open?id=1l7cC58tLhf-nAps3Dzk00FttPPchNaDP (includes all missing sounds and a temporary mix to solve a missing sound issue until done properly)
-Fire Damage Rate
-Speed
-Restored Projector
-How much vehicle is thrown around by weapon fire
-Exit points


HellDragon-HK:
Sigh dang it, I had the driving pose almost perfect on this one, but either the stock model's feet go through the top or my preferred skaarj hybrid model's feet go through the floor, so this last update I found a happy medium which isn't perfect for either but the clipping isn't visible at least and it's not easy to see inside the Scorpion anyway so most probably won't notice how the stock model's legs float while the other goes slightly through front (not far enough to be seen though)

Classes/UT3Scorpion.uc
```diff
@@ -2,6 +2,7 @@
  * Copyright © 2008, 2014 GreatEmerald
  * Copyright © 2008-2009 Wormbo
  * Copyright © 2012 100GPing100
+ * Copyright © 2017 HellDragon (Copyright seems too strong a word to me but if that's what you want)
```
GreatEmerald: No comments in copyright

Classes/UT3Scorpion.uc
```diff
@@ -45,6 +46,14 @@ var IntBox BoostIconCoords, EjectIconCoords;
 var float LastBoostAttempt, SpeedAtBoost;
 var() float MinEjectSpeed;
 var int AirBoost;
+Var rotator FootDriveL,FootDriveR;
+Var rotator ArmDriveL,ArmDriveR;
+Var rotator ForeArmDriveL, ForeArmDriveR;
+Var rotator ThighDriveL,ThighDriveR;
+Var rotator CalfDriveL,CalfDriveR;
+Var rotator SpineDrive;
+Var rotator NeckDrive;
```
GreatEmerald: Again these need to be either local or their properties inside `defaultproperties`

Classes/UT3Scorpion.uc
```diff
@@ -45,6 +46,14 @@ var IntBox BoostIconCoords, EjectIconCoords;
...
+var bool bAttachedDriver;
```
GreatEmerald: Again this is needless

Classes/UT3Scorpion.uc
```diff
@@ -539,12 +548,72 @@ simulated event DrivingStatusChanged()
...
+}
+
 //=============================================================================
 // Default values
 //=============================================================================
 
 defaultproperties
 {
+
+    //DrawScale=1.2
```
GreatEmerald: Again check whether 1.0 is the actual default here

HellDragon-HK:
It is from ONSRV
	DrawScale=1.0
	DrawScale3D=(X=1.0,Y=1.0,Z=1.0)

GreatEmerald:
Note that since you've been working on this during both 2017 and 2018, your copyright in those files you edited during both years should read `2017-2018`.

GreatEmerald:
You'll need to remove the binary upload from this pull request. That probably means that you'll need to redo it anyway and thus it's better to do after the binary purge.

HellDragon-HK:
Redone in new request

PR #89, opened 2017-12-24
UT3 Hoverboard Improvement on Driver position & pose plus camera

HellDragon-HK:
Changed the Driver position to not be floating, adjusted arms to be closer to UT3 (attempted to turn around but no good, animation doesn't fit although that might be fixable with a lot of trial and error, except head faces looking towards the back even while moving forward).

Vehicle mass reduced which slightly helps with the issue of pushing other vehicles so easily
Camera slightly adjusted and extra options added

### DIFF
--- a/Classes/UT3Hoverboard.uc
+++ b/Classes/UT3Hoverboard.uc
@@ -769,16 +769,18 @@ simulated function AttachDriver(Pawn P)
 
     //Arms
     //LeftArm
-    ArmDriveL.Yaw=9000;
-    ArmDriveL.Pitch=-6000;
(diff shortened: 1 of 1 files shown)


GreatEmerald:
Thanks, looks good.

Don't worry about the camera distance: the ability to change it is something UT2004 does better than UT3 anyway.

HellDragon-HK:
Normal cam and Aerial View are very different, in fact the AV cam even has a collision ability like UT3 to some degree as it won't really clip through vehicles, and Alex requested the Gears style cam (kind of) which makes it toward an over the shoulder look

PR #90, opened 2017-12-25
Update UT3ScorpionTurret.uc

HellDragon-HK:
In UT3 Scorpion can't aim straight up, it does have a limit point however it is higher than we can get without clipping and seeing inside or through the Scorpion so this sets it as close as we can get without clipping which is slightly less by maybe 1000-3000 it's not too far off.


GreatEmerald:
Yea, that's fair enough, thanks.

PR #91, opened 2017-12-25
Update UT3CicadaRocket.uc

HellDragon-HK:
Sound fix


HellDragon-HK:
No backup file so doesn't look like it, it's an unused sound fix

GreatEmerald:
OK, thanks.

PR #92, opened 2017-12-25
Update UT3CicadaMissileLauncher.uc

HellDragon-HK:
Sound fix, in UT3 I can only hear the eject sound with alt fire


GreatEmerald:
Same question here, any changes needed to the binaries?

HellDragon-HK:
Doesn't look like it, I think the only sound that was missing was the turret which is pretty much the same as UT2004 and I figured that's why it was left out

GreatEmerald:
OK, thanks.

PR #95, opened 2017-12-25
Update UT3ScorpionBallRed.uc

HellDragon-HK:


GreatEmerald:
Uh, how about the blue one?

HellDragon-HK:
The blue one was done first and is a previous request, do I really need to delete one and add it to both in a single request?

HellDragon-HK:
Remember credit Alex for it, not me

PR #96, opened 2017-12-25
Leviathan shield effect scale

HellDragon-HK:


GreatEmerald:
You should merge #97 with this one, lest they go out of sync.

HellDragon-HK:
How, it says a file of this name already exists use another name

GreatEmerald:
Close the other one, go to your branch (patch-29), edit the other file the same way you did in the other pull request.

GreatEmerald:
Thanks. But just to be sure, are you certain that X should be 0.7 and not 1.0? It doesn't sound right to me.

HellDragon-HK:
Using the screens I got from UT3 of bots using the shields and the old drawscale yes....using the new UT3 drawscale...probably not accurate now but it should be right for the smaller drawscale currently in use

GreatEmerald:
OK, would be nice if someone could confirm (perhaps also update for the UT3 DrawScale values).

HellDragon-HK:
Working on getting UT3 Drawscale updates, all I've got left of Axon is Stealthbender, Hellfire SPMA, and Leviathan to go I believe

GreatEmerald:
Huh, now it's even more skewed compared to the original mesh?..

GreatEmerald:
Also there's an extra `+` in the code...

HellDragon-HK:
What do you mean? In-game it's larger than it was before.

HellDragon-HK:
I see the extra plus and got it I think

btw the Levi shields (specifically Levi) are not actually blocking anything even if I go back to all the original un-edited .uc files.

GreatEmerald:
Since we've determined that the Leviathan shields use a different mesh altogether, it might be better to port that one instead of reusing the Paladin shield mesh. I'm not sure what the difference is, however. Some comparison screenshots of the two could be useful.

HellDragon-HK:
https://github.com/GreatEmerald/UT3Vehicles/pull/74
https://github.com/GreatEmerald/UT3Vehicles/pull/85#issuecomment-354654817

From the screens I already have I'd say the difference is the middle on the Levi shield is open to be able to fire through it unlike the Paladin which rebounds it's shot, don't think there is any other difference

GreatEmerald:
Yea, that's a good point, but also check in the UT3 editor how the two meshes compare (also including size and shape).

Oh yea, and please merge #74 with this pull request, it's literally the same issue.

HellDragon-HK:
Alright then, if I can find them that is, UT3 editor isn't exactly easy to locate things.

HellDragon-HK:
@GreatEmerald Here you go
ss_001
ss_002
ss_003
ss_004

GreatEmerald:
OK, then in the mean while we can use the same mesh for both. I'll add collision notes to the issue.

Then the scale should correspond to the one in UT3: 1.0 here for the Leviathan and 2.2 for the Paladin, on all three axes.

HellDragon-HK:
Where did you find that, using the vehicles drawscale isn't even large enough so how does that work?

GreatEmerald:
I mentioned that here: https://github.com/GreatEmerald/UT3Vehicles/pull/74#issuecomment-354604553

HellDragon-HK:
1.0 looks too small to me but if that's what it's supposed to be....

Still doesn't stop anything however, AVRiLs look like they are going straight through and I was also taking a little health damage while on the laser turret from Raptor plasma

GreatEmerald:
To make sure that the two meshes in UT3 have the same scale, you could also try placing them in a level side by side.

Hm, I'm not seeing where the collision wouldn't match, since these shields extend the Paladin shield. Oh well, we can figure that out later.

HellDragon-HK:
I don't know enough about the editors to do that, UT2004 I might be able to figure it out but UT3 not likely, it's already complaining about 'access denied' with just even looking at things and trying to load a map is a 'you can't edit already cooked files'

GreatEmerald:
I think that warning is about saving, it should still allow you to try things out. Is there a "place static mesh here" if you right-click on a surface in UT3Ed?

HellDragon-HK:
Don't see a 'Place Static Mesh Here' option, there are right click options but nothing direct, I have 0% knowledge of map editing to understand what any of it does

Classes/UT3LeviathanShieldEffectBlue.uc
```diff
@@ -18,7 +18,7 @@ DefaultProperties
         CoordinateSystem=PTCS_Relative
         MaxParticles=1
         UniformSize=false
-        StartSizeRange=(X=(Min=0.6,Max=0.6),Y=(Min=0.75,Max=0.75),Z=(Min=1.0,Max=1.0))
+        StartSizeRange=(X=(Min=0.7,Max=0.7),Y=(Min=1.4,Max=1.4),Z=(Min=1.6,Max=1.6))
```
GreatEmerald: This is out of sync with the rest.

Classes/UT3LeviathanShieldEffectRed.uc
```diff
@@ -18,7 +18,8 @@ DefaultProperties
         CoordinateSystem=PTCS_Relative
         MaxParticles=1
         UniformSize=false
-        StartSizeRange=(X=(Min=0.6,Max=0.6),Y=(Min=0.75,Max=0.75),Z=(Min=1.0,Max=1.0))
+        StartSizeRange=(X=(Min=1.0,Max=1.0),Y=(Min=1.0,Max=1.0),Z=(Min=1.0,Max=1.0))
+        //StartSizeRange=(X=(Min=0.7,Max=0.7),Y=(Min=1.4,Max=1.4),Z=(Min=1.6,Max=1.6))
```
GreatEmerald: No need to keep the previous testing values around as comments.

GreatEmerald:
Or is there a binary requirement to begin with? `UT3A_Vehicle_Paladin.Sounds.A_Vehicle_Paladin_ShieldActivate01` is now used, but that was always there, right?

GreatEmerald:
Hm, you also mentioned the DrawScale update here, so what's the status of that?

Classes/UT3LeviathanShieldEffectBlue.uc
HellDragon-HK: Argh so I did miss one

Classes/UT3LeviathanShieldEffectRed.uc
HellDragon-HK: I'd like to keep that as to me those values are more correct, the 1.0 just seems small but that's just my opinion and possibly an optical illusion as I will say on the hostile side of the Levi shield it doesn't look as big for some reason as it does when you're driving.....odd I agree

HellDragon-HK:
Yeah the only binary here is the ukx

GreatEmerald:
OK, I'll accept this because it's better than what's in master at the moment. We can tweak it further (and figure out what's going on with the collision) later.

PR #98, opened 2017-12-26
Update UT3Paladin.uc

HellDragon-HK:
Gives the Paladin it's compressed state like the Hellbender has


GreatEmerald:
There's a conflict, please rebase on current master.

Also, it would be nice if someone else could test this as well.

HellDragon-HK:
Wouldn't count as resolved unless I replaced the entire thing with everything in my class file, not sure if that's going to be an issue or not

GreatEmerald:
No, this doesn't seem to work, because now it's proposing to revert all changes that 100GPing100 has done so far. Maybe try a new pull request.

HellDragon-HK:
Figured as much ok then

PR #100, opened 2017-12-27
UT3 Hellbender Rear Gun Gunner Pose & Cam

HellDragon-HK:
I did some work getting this pose, hope you like it, once again the arms just refuse to do what I wanted so I got it as best I could, further attempts to improve it did the opposite....also I'm trying to make these positions work with 2 different driving animations, the stock and then my preferred skaarj hybrid model which unfortunately has the legs bent lower (by a lot) and the arms in a different position, the good thing about this animation is it actually fits the stock UT vehicles perfectly the bad it's hard to fit into these UT3 seats without messing up the stock model's pose.

Cams slightly adjusted as well

### DIFF
--- a/Classes/UT3HellbenderRearGunPawn.uc
+++ b/Classes/UT3HellbenderRearGunPawn.uc
@@ -1,5 +1,6 @@
 /*
  * Copyright © 2008, 2014 GreatEmerald
+ * Copyright © 2017 HellDragon (still don't feel right about it but here you go)
  *
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions are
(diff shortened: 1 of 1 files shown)


Classes/UT3HellbenderRearGunPawn.uc
```diff
@@ -1,5 +1,6 @@
 /*
  * Copyright © 2008, 2014 GreatEmerald
+ * Copyright © 2017 HellDragon (still don't feel right about it but here you go)
```
GreatEmerald: Don't put comments into the copyright line.

Classes/UT3HellbenderRearGunPawn.uc
```diff
@@ -39,10 +40,96 @@
 
 class UT3HellbenderRearGunPawn extends ONSPRVRearGunPawn;
 
+Var rotator SpineDrive;
+Var rotator ThighDriveL,ThighDriveR;
+Var rotator CalfDriveL,CalfDriveR;
+Var rotator FootDriveL,FootDriveR;
+Var rotator ArmDriveL,ArmDriveR;
+Var rotator ForeArmDriveL, ForeArmDriveR;
+Var rotator NeckDrive;
```
GreatEmerald: Same comment here: either make these local variables inside the `AttachDriver` function, or...

Classes/UT3HellbenderRearGunPawn.uc
```diff
@@ -39,10 +40,96 @@
...
+var bool bAttachedDriver;
+
+simulated function AttachDriver(Pawn P)
+{
+    super.AttachDriver(P);
+    bAttachedDriver=true;
+
+    NeckDrive.Yaw=4000;
```
GreatEmerald: ...remove these definitions and put them into `defaultproperties` instead.

Classes/UT3HellbenderRearGunPawn.uc
```diff
@@ -39,10 +40,96 @@
...
+    P.SetBoneRotation('Bip01 L ForeArm');
+    P.SetBoneRotation('Bip01 R ForeArm');
+    P.SetBoneRotation('Bip01 L Thigh');
+    P.SetBoneRotation('Bip01 R Thigh');
+    P.SetBoneRotation('Bip01 L Calf');
+    P.SetBoneRotation('Bip01 R Calf');
+    P.SetBoneRotation('Bip01 L Foot');
+    P.SetBoneRotation('Bip01 R Foot');
+    
+    bAttachedDriver=false;
```
GreatEmerald: Remove these as well, this variable is unused.

Classes/UT3HellbenderRearGunPawn.uc
HellDragon-HK: Doesn't work, it won't compile that way

Classes/UT3HellbenderRearGunPawn.uc
HellDragon-HK: C:\Unreal Anthology\UT2004\UT3Vehicles\Classes\UT3HellbenderRearGunPawn.uc(48) : Error, 'Var' is not allowed here

GreatEmerald: Make them `local` if you want to go that way.

Classes/UT3HellbenderRearGunPawn.uc
GreatEmerald: It will as long as it is in the defaultproperties rotator form of NeckDrive=(Yaw=4000) etc.

GreatEmerald:
OK, thanks.

PR #101, opened 2017-12-27
Update UT3GoliathTurret.uc

HellDragon-HK:
Ok let's take it from the top and try to do this right shall we...

Pitch adjusted to be closer to UT3, weapon fire adjusted to be closer to weapon barrel and not as floaty, scaling damage type was needed and can be found here https://drive.google.com/open?id=11uMpAiP9SUgkJWbzIhwQwzvTh87PMDNc , spread and damage changed to UT3 values, aim error UT3 value added.

### DIFF
--- a/Classes/UT3DmgType_GoliathMachineGun.uc
+++ b/Classes/UT3DmgType_GoliathMachineGun.uc
@@ -46,6 +46,6 @@ defaultproperties
     MaleSuicide="%o turned the minigun on himself."
     FemaleSuicide="%o turned the minigun on herself."
     KDamageImpulse=700 //2000 def UT2004, 700 def UT3
-    VehicleDamageScaling=0.6  //0.36 def UT2004, 0.5 def UT3
+    VehicleDamageScaling=0.5  //0.36 def UT2004, 0.5 def UT3 but does 9 damage, only 8 in UT2004
     VehicleClass=class'UT3GoliathTurretPawn'
 }
--- a/Classes/UT3DmgType_GoliathMachineGun.uc
+++ b/Classes/UT3DmgType_GoliathMachineGun.uc
@@ -1,0 +1,53 @@
+/*
+ * Copyright © 2014 GreatEmerald
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are
+ * met:
+ *
+ *     (1) Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *
+ *     (2) Redistributions in binary form must reproduce the above copyright
+ *     notice, this list of conditions and the following disclaimers in
+ *     the documentation and/or other materials provided with the
+ *     distribution.
+ *
+ *     (3) The name of the author may not be used to
+ *     endorse or promote products derived from this software without
+ *     specific prior written permission.
+ *
+ *     (4) The use, modification and redistribution of this software must
+ *     be made in compliance with the additional terms and restrictions
+ *     provided by the Unreal Tournament 2004 End User License Agreement.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE AUTHOR ``AS IS'' AND ANY EXPRESS OR
+ * IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR ANY DIRECT,
+ * INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+ * STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING
+ * IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ *
+ * This software is not supported by Atari, S.A., Epic Games, Inc. or any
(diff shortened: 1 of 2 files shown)


Classes/UT3DmgType_GoliathMachineGun.uc
```diff
@@ -0,0 +1,53 @@
+/*
+ * Copyright © 2014 GreatEmerald
```
GreatEmerald: Your copyright and year

Classes/UT3DmgType_GoliathMachineGun.uc
```diff
@@ -0,0 +1,53 @@
...
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+ * STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING
+ * IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ *
+ * This software is not supported by Atari, S.A., Epic Games, Inc. or any
+ * of such parties' affiliates and subsidiaries.
+ */
+
+class UT3DmgType_GoliathMachineGun extends VehicleDamageType;
```
GreatEmerald: Extend the UT2004 Goliath minigun DamageType

Classes/UT3DmgType_GoliathMachineGun.uc
```diff
@@ -0,0 +1,53 @@
...
+
+defaultproperties
+{
+    bRagdollBullet=true
+    DeathString="%o was punctured by %k's minigun."
+    MaleSuicide="%o turned the minigun on himself."
+    FemaleSuicide="%o turned the minigun on herself."
```
GreatEmerald: Make sure these are set to what it is in UT3

Classes/UT3DmgType_GoliathMachineGun.uc
```diff
@@ -0,0 +1,53 @@
...
+    FlashFog=(X=600.00000,Y=0.000000,Z=0.00000)
+    KDamageImpulse=700 //2000 def UT2004, 700 def UT3
+    VehicleDamageScaling=0.6  //0.36 def UT2004, 0.5 def UT3
```
GreatEmerald: If it's 0.5 by default in UT3, then why is it set to 0.6 here?

Classes/UT3GoliathTurret.uc
```diff
@@ -2,7 +2,8 @@
 // UT3GoliathTurret.uc
 // I don't think the sounds will initially work...
 // GreatEmerald, 2008, 2014
-// Copyright (c) 2012, 100GPing100 (visuals + fixed sounds)
+// Copyright © 2012, 2017 Luís 'zeluisping' Guimarães <zeluis.100@gmail.com> (visuals + fixed sounds)
```
GreatEmerald: @100GPing100 Do you actually want your email to be in the files? That can't be good for spam protection...

Classes/UT3GoliathTurret.uc
```diff
@@ -2,7 +2,8 @@
...
+// HellDragon 2017 (I guess it helps identify anything I changed if there's a question about it later)
```
GreatEmerald: You're missing `Copyright ©`, and no, you shouldn't actually list what you changed in the copyright line. Git keeps track of that instead. In case something goes wrong, we can `git blame` you ;)

Classes/UT3DmgType_GoliathMachineGun.uc
HellDragon-HK: .5 wasn't giving the same amount of damage, 0.6 gives exactly the same damage, no idea why

Classes/UT3GoliathTurret.uc
HellDragon-HK: Am I supposed to try to change that for GP, I did it here because that's what he did with the Paladin and assumed it would make things easier

Classes/UT3GoliathTurret.uc
HellDragon-HK: Oh....goody.....although I'll admit when something goes wrong I check to see if it was my fault and if yes I'm the first to say so usually

Classes/UT3DmgType_GoliathMachineGun.uc
HellDragon-HK: If I recall, I was only getting 8 damage, in UT3 it's 9, with 0.6 I got that

Classes/UT3GoliathTurret.uc
GreatEmerald: Let's wait for is reply.

Classes/UT3GoliathTurret.uc
GreatEmerald: You still haven't fixed this issue though!

Classes/UT3DmgType_GoliathMachineGun.uc
GreatEmerald: That's not good enough. If it's 0.5 in UT3, it must be 0.5 here as well. If it doesn't match what you see, we need to figure out why instead of changing numbers.

HellDragon-HK: Open to suggestions because I didn't see anything else

GreatEmerald: I see nothing in the code to suggest that it's 9 damage in UT3.

HellDragon-HK: That's what I get in-game on everything but I'll double check when I can, trying to find out about Levi turret throwing shards or not atm

Classes/UT3GoliathTurret.uc
100GPing100: Well, I don't have a problem with it, but just in case, accept the pull request with that and I'll submit a pull request to remove all references to my e-mail and replace it with something else.

Classes/UT3DmgType_GoliathMachineGun.uc
HellDragon-HK: I'm getting 9 damage in UT3 on all vehicles on VCTF-Suspense with a single shot

Classes/UT3GoliathTurret.uc
GreatEmerald: OK.

Classes/UT3DmgType_GoliathMachineGun.uc
GreatEmerald: Well, put this on 0.5 anyway. There's a number of reasons why there may be a one point difference and `VehicleDamageScaling` is most certainly not it. We can figure it out later.

GreatEmerald:
OK, thanks.

PR #102, opened 2017-12-27
Update UT3LeviathanBolt.uc

HellDragon-HK:


Classes/UT3DmgType_LeviathanBolt.uc
```diff
@@ -0,0 +1,56 @@
+/*
+ * Copyright © 2014 GreatEmerald
```
GreatEmerald: Your copyright

Classes/UT3DmgType_LeviathanBolt.uc
```diff
@@ -0,0 +1,56 @@
...
+ * of such parties' affiliates and subsidiaries.
+ */
+
+class UT3DmgType_LeviathanBolt extends VehicleDamageType;
+
+defaultproperties
+{
+    DeathString="%o rode %k's rocket into oblivion."
+    MaleSuicide="%o fired his rocket prematurely."
+    FemaleSuicide="%o fired her rocket prematurely."
```
GreatEmerald: These messages are almost certainly not right

Classes/UT3DmgType_LeviathanBolt.uc
```diff
@@ -0,0 +1,56 @@
...
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+ * STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING
+ * IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ *
+ * This software is not supported by Atari, S.A., Epic Games, Inc. or any
+ * of such parties' affiliates and subsidiaries.
+ */
+
+class UT3DmgType_LeviathanBolt extends VehicleDamageType;
```
GreatEmerald: Extend the UT2004 version instead

Classes/UT3LeviathanBolt.uc
```diff
@@ -34,7 +35,8 @@ defaultproperties
 	Damage=100
 	DamageRadius=300
 	MomentumTransfer=4000
-	
+        LifeSpan=4.0
```
GreatEmerald: There is one space too much here

Classes/UT3DmgType_LeviathanBolt.uc
HellDragon-HK: What do you mean here, DamTypeRocket won't compile and according to the death message it's plasma not rockets so extending as the Levi rocket actually wouldn't be right either, it actually needs to be Levi Turret or Manta

Classes/UT3LeviathanBolt.uc
HellDragon-HK: The extra space isn't there on edit nor my class file and I can't re save it here without making some kind of change :S

Classes/UT3DmgType_LeviathanBolt.uc
GreatEmerald: Yes, the Leviathan turret. Its default properties are probably not going to be very useful, but it might help with compatibility with other mods that might check whether the projectile comes from a Leviathan by looking at the `DamageType`.

Classes/UT3LeviathanBolt.uc
GreatEmerald: Try removing one space from the ones that are there, perhaps?..

Classes/UT3DmgType_LeviathanBolt.uc
HellDragon-HK: By Levi turret I meant the side turrets since they fire plasma by default, which one do you mean?

GreatEmerald: The vanilla UT2004 one.

HellDragon-HK: Yes but which vanilla one, the RocketPack or SideGun?

GreatEmerald: RocketPack. SideGuns are the passenger weapons and the bolt is from the main gun.

HellDragon-HK:
That would need to extend as DamTypeRocket though and it won't compile

GreatEmerald:
Oh, you're right. That's really weird, I guess Epic were lazy here too. All right, otherwise this looks fine.

GreatEmerald:
Wait a minute, the death strings still don't seem right?.. This is not plasma, it's a stinger shard...

HellDragon-HK:
Nope it's plasma according to UTGameContent.int

[UTDmgType_LeviathanBolt]
DeathString="`k's Leviathan filled `o with plasma."
FemaleSuicide="`o fried herself with her own plasma blast."
MaleSuicide="`o fried himself with his own plasma blast."

GreatEmerald:
Well, the `DeathString` doesn't quite match what you posted here, in any case.

HellDragon-HK:
Oops it's backward lol

PR #103, opened 2017-12-27
Update UT3LeviathanPrimaryWeapon.uc

HellDragon-HK:


Classes/UT3DmgType_LeviathanCannon.uc
```diff
@@ -0,0 +1,61 @@
+/*
+ * Copyright © 2014 GreatEmerald
```
GreatEmerald: This should be you and the current date.

Classes/UT3DmgType_LeviathanCannon.uc
```diff
@@ -0,0 +1,61 @@
...
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+ * STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING
+ * IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ *
+ * This software is not supported by Atari, S.A., Epic Games, Inc. or any
+ * of such parties' affiliates and subsidiaries.
+ */
+
+class UT3DmgType_LeviathanCannon extends VehicleDamageType;
```
GreatEmerald: This should rather extend `DamTypeMASCannon`. The closer to stock UT2004 classes, the better.

Classes/UT3DmgType_LeviathanCannon.uc
```diff
@@ -0,0 +1,61 @@
...
+
+defaultproperties
+{
+    DeathString="%o was VAPORIZED by %k!"
+    MaleSuicide="%o was VAPORIZED!"
+    FemaleSuicide="%o was VAPORIZED!"
```
GreatEmerald: Are these strings from UT3?

Classes/UT3DmgType_LeviathanCannon.uc
```diff
@@ -0,0 +1,61 @@
...
+    VehicleDamageScaling=1.5
+    
+    bArmorStops=False
+    bNeverSevers=True
```
GreatEmerald: What are these for?

HellDragon-HK:
"Another comment is that you should integrate damage amount changes into this code too, as I mentioned in issue #72."

I would but I don't know how to change that correctly, looks to be very differently done unless you mean in the default properties the min and max, that I can do.

Classes/UT3DmgType_LeviathanCannon.uc
HellDragon-HK: Default UT2004 MAS Cannon Dmg Type and no clue

GreatEmerald:
You just need to override the state and the function inside and change that one line. This ought to do (though it needs tabs to spaces):
```UnrealScript
simulated state InstantFireMode
{
    function Explosion(float DamRad)
	{
    	local actor Victims;
    	local float damageScale, dist;
    	local vector Dir;

    	if (Role < ROLE_Authority)
    	   return;

    	foreach VisibleCollidingActors(class 'Actor', Victims, DamRad, GHitLocation)
    	{
    		if( (Victims != self) && (Victims != Instigator) && (Victims.Role == ROLE_Authority) && (!Victims.IsA('FluidSurfaceInfo')) )
    		{
    			Dir = Victims.Location - GHitLocation;
     			dist = FMax(1,VSize(Dir));
    			Dir = Dir/dist;
    			Dir.Z *= 5.0;
    			Dir = Normal(Dir);
    			damageScale = 1;

                if (Pawn(Victims) != None && Pawn(Victims).GetTeamNum() == Instigator.GetTeamNum())
                    damageScale = 0;

    			Victims.TakeDamage(
                    				damageScale * DamageMax,
                    				Instigator,
                    				Victims.Location - 0.5 * (Victims.CollisionHeight + Victims.CollisionRadius) * Dir,
                    				(damageScale * Momentum * Dir),
                    				DamageType
                    			  );
    		}
    	}
	}
}
}
```

If the `simulated state InstantFireMode` is already overridden, then just add the `function Explosion` override inside it.

HellDragon-HK:
Didn't help, the core was destroyed in a single long range shot

GreatEmerald:
Did you also make sure to set damage to the UT3 values? Either way submit that to this pull request.

HellDragon-HK:
What's the values though this is all I can find is this it?

	ExplosionDamage=150.0
	ExplosionRadius=1000.0
	ExplosionMomentum=125000

HellDragon-HK:
Wait I got it, forgot you already said the real values were the EmitExplosion which has 250 and using that it now takes 3 shots with the first bringing it down to 58%....so then what are the other values?

GreatEmerald:
I'm not sure where you even got those other values from. Indeed the emitter is what causes all the damage in UT3.

You didn't apply the correct indentation yet, though, and there are still tabs that need to be converted into spaces.

HellDragon-HK:
Found those values here http://greatemerald.xmpcommunity.com/uncodex/UT3-15/utgamecontent/utvehicle_leviathan_content.html  so what are they then?

GreatEmerald:
Oh, that's the damage it deals when the vehicle itself explodes (according to the comments in the source, when it's upside-down, but maybe it explodes on death too?).

The indentation is still not quite right.

GreatEmerald:
Well, I guess the indentation can be fixed later...

PR #107, opened 2017-12-27
Update UT3LeviathanTurretWeaponMinigun.uc

HellDragon-HK:


Classes/UT3DmgType_LeviathanShard.uc
```diff
@@ -0,0 +1,54 @@
+/*
+ * Copyright © 2014 GreatEmerald
```
GreatEmerald: Fix this

Classes/UT3DmgType_LeviathanShard.uc
```diff
@@ -0,0 +1,54 @@
...
+ */
+
+class UT3DmgType_LeviathanShard extends VehicleDamageType;
+
+defaultproperties
+{
+    bRagdollBullet=true
+    DeathString="%o was punctured by %k's minigun."
+    MaleSuicide="%o turned the minigun on himself."
+    FemaleSuicide="%o turned the minigun on herself."
```
GreatEmerald: These are almost certainly not right.

Classes/UT3DmgType_LeviathanShard.uc
```diff
@@ -0,0 +1,54 @@
...
+    FlashFog=(X=600.00000,Y=0.000000,Z=0.00000)
+    KDamageImpulse=2000
+    VehicleMomentumScaling=0.5
+    VehicleDamageScaling=0.5  //0.36 def UT2004, 0.5 def UT3
+    VehicleClass=class'UT3LeviathanTurretStinger'
+    bBulletHit=True
```
GreatEmerald: Values should be based on the assumption that it should be shard-based rather than bullet-based. So `bBulletHit` and `bRagdollBullet` are not applicable.

GreatEmerald:
Not really, stick with the vanilla wording even if it's not very exciting. We don't even know what the official name of the turret is, after all.

GreatEmerald:
This depends on #119.

HellDragon-HK:
Doesn't fire shards in either PC or PS3, just shock beams with stinger sound, the red Killed By message says Stinger Turret

GreatEmerald:
OK, then you do need to put back `bBulletHit` etc. into the DamageType again since it's basically a reskinned minigun.

HellDragon-HK:
Any chance of doing it right and not the lazy way like Epic went (which I still think was changed in a patch somewhere possibly for performance on PS3)

GreatEmerald:
I meant that you should readd the changes you had in this commit: https://github.com/GreatEmerald/UT3Vehicles/pull/107/commits/d8f5d65c3b5d5a5212d139b1c2d87957249cb127

And yes, ideally the stinger turret would use the Stinger effects from the weapons. But we don't have Stinger effects for the weapons either, so that's for some time later.

Classes/UT3DmgType_LeviathanShard.uc
```diff
@@ -0,0 +1,55 @@
...
+    bRagdollBullet=true
+    DeathString="%k speared %o."
+    MaleSuicide="%o died."
+    FemaleSuicide="%o died."
+    FlashFog=(X=600.00000,Y=0.000000,Z=0.00000)
+    KDamageImpulse=2000
+    VehicleMomentumScaling=0.5
+    VehicleDamageScaling=0.5  //0.36 def UT2004, 0.5 def UT3
+    VehicleClass=class'UT3LeviathanTurretStinger'
+    bBulletHit
```
GreatEmerald: This needs to be `bBulletHit=True`, as it is it won't even compile...

HellDragon-HK:
Huh how what happened to the rest of it surely I wouldn't have made that dumb of a mistake

GreatEmerald:
OK, thanks.

Issue #110, opened 2017-12-27
Team-coloured headlight coronas


GreatEmerald:
In UT3, headlights are team-coloured. In UT2004, they are team-coloured too, but you wouldn't know if you didn't look real closely at them! This effect should be more pronounced.

In UT2004, this is handled by the ONSHeadlightCorona class: http://greatemerald.xmpcommunity.com/uncodex/UT2004/Source_onslaught/onsheadlightcorona.html#4 However, these values are hardcoded, and the class of the headlight coronas is hardcoded too, which is a problem.

Several approaches could be used in this case:

- In each vehicle, replace the bit of code that spawns the `ONSHeadlightCorona` with one that spawns our custom `UT3HeadlightCorona` with the correct values;
- Use `CheckReplacement()` in a mutator to replace all `ONSHeadlightCorona`s with `UT3HeadlightCorona`s;
- Add new code to each vehicle to set the hue and saturation manually after each team switch (`TeamChanged`).

One big difference between these approaches would be network compatibility. The default headlight has `RemoteRole=ROLE_None`, which means it is completely client-side. I'm not sure if destroying it would be replicated or not... If not, you'd get two headlights on top of one another when using approach 2. Approach 3 would work as long as the code is run by the clients as well (it should be). Approach 1 would always work, but it requires copy-pasting a lot of code to each vehicle.

I'd probably go for option 3, and if it doesn't work, for option 1.

HellDragon-HK:
Speaking of UT3HeadlightCorona, here is the uc I was trying to use https://drive.google.com/open?id=1BBPpls85AuQW8tSwLjAY5y3yEPo_09NE basically a direct copy of the original but with Alex's color suggestions

"In each vehicle, replace the bit of code that spawns the ONSHeadlightCorona with one that spawns our custom UT3HeadlightCorona with the correct values;"

This is what I tried, and the effect works but it's always turned on even when empty and doesn't disappear on death either

GreatEmerald:
That is strange, it ought to work, because UT3HeadlightCorona is an ONSHeadlightCorona (so it should be spawned and destroyed correctly with the existing code). You should put this up as a pull request so I can check for any things that may be missing.

HellDragon-HK:
This class or the vehicle I tried to change, which I don't have the full changes anymore as it went badly and I reverted but I think I download the test file I sent Alex of where it worked without adversely affecting the vehicle

HellDragon-HK:
[...]

PR #112, opened 2017-12-27
Manta Headlight Corona Examination

HellDragon-HK:
DO NOT MERGE THIS, it's the one you wanted to look at for the headlight corona not disappearing issue

### DIFF
--- a/Classes/UT3HeadlightCorona.uc
+++ b/Classes/UT3HeadlightCorona.uc
@@ -1,0 +1,79 @@
+/*
+ * Copyright © 2014 GreatEmerald
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are
+ * met:
+ *
+ *     (1) Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *
+ *     (2) Redistributions in binary form must reproduce the above copyright
+ *     notice, this list of conditions and the following disclaimers in
+ *     the documentation and/or other materials provided with the
+ *     distribution.
+ *
+ *     (3) The name of the author may not be used to
+ *     endorse or promote products derived from this software without
+ *     specific prior written permission.
+ *
+ *     (4) The use, modification and redistribution of this software must
+ *     be made in compliance with the additional terms and restrictions
+ *     provided by the Unreal Tournament 2004 End User License Agreement.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE AUTHOR ``AS IS'' AND ANY EXPRESS OR
+ * IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR ANY DIRECT,
+ * INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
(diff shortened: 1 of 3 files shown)


GreatEmerald:
Ah yes, I see what causes the behaviour you were talking about. You have in `PostNetBeginPlay` a call to `Super.PostNetBeginPlay()`. This calls the `PostNetBeginPlay` function of the parent, `ONSVehicle`, which includes code to spawn an `ONSHeadlightCorona`! Which means that first an `ONSHeadlightCorona` gets spawned, then your code spawns a `UT3HeadlightCorona` *on top* of the `ONSHeadlightCorona`! And when the vehicle is destroyed, it destroys the `UT3HeadlightCorona`, but it loses track of the `ONSHeadlightCorona`, which is what you see.

The solution is to copy all `PostNetBeginPlay` code from http://greatemerald.xmpcommunity.com/uncodex/UT2004/Source_onslaught/onsvehicle.html#452 (yes, that's a lot of code), replace the first `Super.PostNetBeginPlay();` with `Super(SVehicle).PostNetBeginPlay();` (which calls the function in `SVehicle`, rather than `ONSVehicle`) and replace the reference to `ONSHeadlightCorona` with `UT3HeadlightCorona`.

HellDragon-HK:
Including the spawn weapons part?

GreatEmerald:
Yes, because there's no way to replace just one line, have to replace them all.

(This issue links to issue #110)

HellDragon-HK:
Same issue, it works when getting in and out but when empty I'm seeing I guess the default corona always on from initial spawn, at least now it's disappearing with the vehicle corpse instead of just staying there endlessly.

HellDragon-HK:
Ok with SVehicle that works but Manta has lost it's hover ability and just scrapes along the ground now but it hasn't lost it weapons, fire rate, or turning ability so that's a step up from what I got before.

GreatEmerald:
Oh yes, that's because `ONSHoverCraft` also has code in `PostNetBeginPlay`. You also have to copy over all this code and put it at the top of the function: http://greatemerald.xmpcommunity.com/uncodex/UT2004/Source_onslaught/onshovercraft.html#114 (note: the "#114" in the URL is a reference to the line number).

HellDragon-HK:
Well it can jump now but still doesn't hover

GreatEmerald:
Put this up in the pull request so I could check.

GreatEmerald:
Hm, what if you move `Super(SVehicle).PostNetBeginPlay();` right before `if (Role == ROLE_Authority)`?

HellDragon-HK:
Hey that seems to have worked and here is what we get from all this hard work although placement seems off for some reason
shot00629
shot00630

GreatEmerald:
Hm, good to know. However, doing all this for every single vehicle is going to be a nightmare.

Putting code to manually set the hue and saturation of headlights on each team change might be a more reasonable approach. For that matter, headlights are completely irrelevant to gameplay, so I'm pretty sure they don't even exist on servers, which should make it easy enough to change the hue by simply using a function like this:

```UnrealScript
simulated function TeamChanged()
{
    Super.TeamChanged();
    if (Level.NetMode != NM_DedicatedServer)
    {
        for(i = 0; i < HeadlightCorona.Length; i++)
        {
            HeadlightCorona[i].LightSaturation = 0;
            if (Team == 0)
                HeadlightCorona[i].LightHue = 0;
            if (Team == 1)
                HeadlightCorona[i].LightHue = 175;
        }
    }
}
```

HellDragon-HK:
They show up on LAN game including projectors, so try that on a different vehicle then maybe because I don't want to undo all the work on the Manta just to try a new approach

HellDragon-HK:
WOW! The red one looks evil with that, I'd say that solution might work, didn't test blowing it up or anything though.

shot00631
shot00633

GreatEmerald:
Yeap, that function should be universal for every vehicle. If it works in LAN games, then that's good enough!

HellDragon-HK:
I meant the default coronas and projectors, I'd need to check if that change works on LAN

HellDragon-HK:
Yeah they show up on both LAN and joining a dedicated LAN

GreatEmerald:
OK, great, then simply put that function in every single vehicle that has headlight coronas.

PR #135, opened 2017-12-28
Raptor Changes for UT3 Drawscale

HellDragon-HK:
-FX placement
-Entry Radius
-Exit Points, including new frontal one
-FP Camera

### DIFF
--- a/Classes/UT3Raptor.uc
+++ b/Classes/UT3Raptor.uc
@@ -295,6 +295,9 @@ Test 4: 275 works well, although it still gives slight sinkness. But that's OK.
 
 defaultproperties
 {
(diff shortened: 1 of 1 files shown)


GreatEmerald:
Does this require binary changes? Is the DrawScale set to 1.0 now?

HellDragon-HK:
Drawscale is 1.3 if I recall and yes because I had to increase the collision box to include the wings initially

GreatEmerald:
As in, in the code. You commented out `DrawScale`, but for it to match what is set in the model (if you scaled the model by the UT3 amount) it should be set to 1.0.

PR #136, opened 2017-12-28
Update UT3RaptorWeapon.uc

HellDragon-HK:
Got the plasma fire a lot closer to the gun barrels now


GreatEmerald:
Looks nice, but I'd like to have someone else confirm that this works as expected.

Unre-Alex:
If this is for the drawscale 1.0 then the projectiles come straight from the gun barrels.

GreatEmerald:
OK, thanks.

HellDragon-HK:
Apparently it works on both drawscales without needing to change, then again I'm finding most weapons to be this way

PR #137, opened 2017-12-28
Cicada Changes for UT3 Drawscale

HellDragon-HK:
Updated for UT3 Drawscale, and added some things I was still working on that was meant to be added to the overall but wasn't ready

-Entry Radius
-Entry Point to place entry more toward front and less on the engines
-Exit Points
-FX placement
-FP Camera, 2 options, 1 in front of Cicada like Raptor and another that has it on right side rocket launcher
-TP Camera


GreatEmerald:
OK, thanks.

PR #139, opened 2017-12-28
Goliath Changes for UT3 Drawscale

HellDragon-HK:
-Entry Radius reduced to to avoid getting into the wrong vehicle
-Exit points
-FX placement
-FP Camera


GreatEmerald:
OK, thanks.

PR #140, opened 2017-12-28
Update UT3GoliathTurretPawn.uc

HellDragon-HK:
FP Camera fix for UT3 Drawscale


GreatEmerald:
Does this depend on binary file updates?

HellDragon-HK:
Yes, the Goliath will have little to no collision if it's not updated and this is the vehicle I checked to see if code drawscale also affected collision which is doesn't

GreatEmerald:
OK, thanks.

PR #145, opened 2017-12-28
Update UT3Fury.uc

HellDragon-HK:


GreatEmerald:
Woah, why is this plasma if the Fury doesn't fire plasma?

HellDragon-HK:
It doesn't have a beam yet so calling it a sweeping beam isn't right and the Fury did have a bolt type damage

[UTDmgType_FuryBolt]
DeathString="`k's Fury filled `o with plasma."
FemaleSuicide="`o fried herself with her own plasma blast."
MaleSuicide="`o fried himself with his own plasma blast."


[UTDmgType_FuryBeam]
DeathString="`o was disintegrated by `k's beam."
MaleSuicide="`o disintegrated himself."
FemaleSuicide="`o disintegrated herself."

GreatEmerald:
Eh, I don't see much reason to merge this as-is, it won't really help in the long run.

HellDragon-HK:
Can't be used for setting up a template to work with later so the Fury is already different from Raptor and less work, otherwise won't we just have to re do this?

GreatEmerald:
We'd have to redo this anyway, because none of the values will match the final ones.

PR #147, opened 2017-12-28
Update UT3PaladinProjectile.uc

HellDragon-HK:
Adds the missing Momentum Scaling

### DIFF
--- a/Classes/UT3PaladinProjectile.uc
+++ b/Classes/UT3PaladinProjectile.uc
@@ -6,6 +6,25 @@
 //-----------------------------------------------------------
 class UT3PaladinProjectile extends ONSShockTankProjectile;
 
+function SuperExplosion()
+{
+	local actor HitActor;
+	local vector HitLocation, HitNormal;
+
+	HurtRadius(ComboDamage, ComboRadius, class'UT3DmgType_PaladinShockBall', ComboMomentumTransfer, Location );
+
+	Spawn(class'ONSShockTankShockExplosion');
+	if ( (Level.NetMode != NM_DedicatedServer) && EffectIsRelevant(Location,false) )
+	{
+		HitActor = Trace(HitLocation, HitNormal,Location - Vect(0,0,120), Location,false);
+		if ( HitActor != None )
+			Spawn(class'ComboDecal',self,,HitLocation, rotator(vect(0,0,-1)));
+	}
+	PlaySound(ComboSound, SLOT_None,1.0,,800);
+    DestroyTrails();
+    Destroy();
+}
+
 DefaultProperties
 {
     Damage=200.000000             //GE: Everything increased!!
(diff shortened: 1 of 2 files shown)


Classes/UT3PaladinProjectile.uc
```diff
@@ -6,6 +6,25 @@
...
+	HurtRadius(ComboDamage, ComboRadius, class'UT3DmgType_PaladinShockBall', ComboMomentumTransfer, Location );
+
+	Spawn(class'ONSShockTankShockExplosion');
+	if ( (Level.NetMode != NM_DedicatedServer) && EffectIsRelevant(Location,false) )
+	{
+		HitActor = Trace(HitLocation, HitNormal,Location - Vect(0,0,120), Location,false);
+		if ( HitActor != None )
+			Spawn(class'ComboDecal',self,,HitLocation, rotator(vect(0,0,-1)));
+	}
+	PlaySound(ComboSound, SLOT_None,1.0,,800);
```
GreatEmerald: This all needs tabs to spaces.

Classes/UT3PaladinProjectile.uc
```diff
@@ -6,6 +6,25 @@
 //-----------------------------------------------------------
 class UT3PaladinProjectile extends ONSShockTankProjectile;
 
+function SuperExplosion()
+{
+        local actor HitActor;
+        local vector HitLocation, HitNormal;
+
+        HurtRadius(ComboDamage, ComboRadius, class'UT3DmgType_PaladinShockBall', ComboMomentumTransfer, Location );
```
GreatEmerald: You can also add a comment here saying that this is the only line that is different from the original.

Classes/UT3PaladinProjectile.uc
```diff
@@ -6,6 +6,25 @@
...
+
+        Spawn(class'ONSShockTankShockExplosion');
+        if ( (Level.NetMode != NM_DedicatedServer) && EffectIsRelevant(Location,false) )
+        {
+            HitActor = Trace(HitLocation, HitNormal,Location - Vect(0,0,120), Location,false);
+	    if ( HitActor != None )
+	        Spawn(class'ComboDecal',self,,HitLocation, rotator(vect(0,0,-1)));
+        }
+        PlaySound(ComboSound, SLOT_None,1.0,,800);
```
GreatEmerald: All of these should be aligned with the lines below, four spaces for each indent. `HitActor` above should be aligned with the `if`.

GreatEmerald:
Still too many spaces starting from the `local actor HitActor;` line, there are 8 and there should be 4, like with the very last lines.

GreatEmerald:
OK, and now since `HitActor = Trace(HitLocation, HitNormal,Location - Vect(0,0,120), Location,false);` and the two lines below it are nested inside the `if`, they should have 8 spaces in front.

GreatEmerald:
...and then `Spawn(class'ComboDecal',self,,HitLocation, rotator(vect(0,0,-1)));` is by itself nested inside the above `if` statement, so it should have additional 4 spaces.

GreatEmerald:
That's the wrong `Spawn`...
The point of this exercise is that you get how nesting works and how statements should be indented. Each `{` increases the indent, each `}` decreases it, and `if` and `for` also increase the indent for a single line after them if there is no `{`.

GreatEmerald:
OK, thanks.

PR #148, opened 2017-12-29
Update UT3HellbenderSideGun.uc

HellDragon-HK:
Drawscale disabled in code to not double with ukx & death message fix kind of

EDIT - Ah got it wrong, apparently SPMA and Hellbender turrets are not the same on message?

[UTDmgType_VehicleShockBeam]   - This is Hellbender as far as I can tell
DeathString="`k's laser shocked `o."
MaleSuicide="`o used his laser on himself."
FemaleSuicide="`o used her laser on herself."

[UTDmgType_SPMAShockBeam]
DeathString="`o was disintegrated by `k's beam."
MaleSuicide="`o disintegrated himself."
FemaleSuicide="`o disintegrated herself."

### DIFF
--- a/Classes/UT3DmgType_HellbenderLaser.uc
+++ b/Classes/UT3DmgType_HellbenderLaser.uc
@@ -4,7 +4,7 @@ class UT3DmgType_HellbenderLaser extends DamTypePRVLaser
 defaultproperties
 {
     VehicleClass=Class'UT3Hellbender'
-    DeathString="%k's laser shocked %o."
+    DeathString="%k's Hellbender laser shocked %o."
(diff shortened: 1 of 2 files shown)


GreatEmerald:
Unless the default DrawScale is 1.0, that line should be left as is...

Classes/UT3DmgType_HellbenderLaser.uc
```diff
@@ -0,0 +1,11 @@
+class UT3DmgType_HellbenderLaser extends DamTypePRVLaser
```
GreatEmerald: This file needs the license header.

GreatEmerald:
OK, thanks.

PR #149, opened 2017-12-29
Update UT3HellfireSPMA.uc

HellDragon-HK:


GreatEmerald:
Does this require an update to binaries?

HellDragon-HK:
Technically no but lights will be floating as well as a bit of a far exit point so yes ukx required

GreatEmerald:
OK, thanks.

PR #150, opened 2017-12-29
Update UT3Leviathan.uc

HellDragon-HK:
Updated for UT3 Drawscale, ukx needs to be updated for this update to be right


GreatEmerald:
OK, thanks.

PR #153, opened 2017-12-30
Update UT3DmgType_ViperBolt.uc

HellDragon-HK:
At any rate, I told you I admit my own mistakes and did with this even though I could have easily let it slip through as an improvement since the vanilla doesn't really match in my opinion


GreatEmerald:
OK. Though this might make 100GPing100 need to do a bit of a rebase since he's working on his own improvements...

PR #154, opened 2017-12-30
Update UT3LeviathanTurretWeapon.uc

HellDragon-HK:
Set a limit of upward aim to help with camera issues, in UT3 turrets can (or seem to) aim straight up, this sets it just a little less where they can almost aim straight up but not quite


GreatEmerald:
OK, looks good.

PR #158, opened 2017-12-30
Goliath: fix stop time

100GPing100:
Fixed issue from #5: make the goliath come to a stop faster.

### DIFF
--- a/Classes/UT3Goliath.uc
+++ b/Classes/UT3Goliath.uc
@@ -1,6 +1,6 @@
 /*
  * Copyright © 2008 Wormbo
- * Copyright © 2012 100GPing100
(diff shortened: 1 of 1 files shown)


GreatEmerald:
Looks nice, thanks.
Again a question I had before: are you sure you want to have your unmasked email in the code? Spambots might parse it rather easily...

100GPing100:
@GreatEmerald You can leave it as is for now, I'll submit a pull request later to modify them all.

GreatEmerald:
OK, but looks like this will need a rebase in any case...

100GPing100:
Yes, I'll probably have to do it like I did with the manta branch and re-do it, would have done it yesterday but wasted a lot of time trying to rebase it properly and failed. I'll also take advantage of that to remove the e-mail from the copyright.

PR #160, opened 2017-12-31
Manta: Engine Pitch & Stop Time

100GPing100:
Fixes issues from #9:

- No longer takes too long to come to a stop
- Reduced engine pitch


HellDragon-HK:
I think the sound is closer now, still seems like it has a lower pitch in UT3 but speed sounds like it might be right, didn't make the uc edits to test stop time yet

GreatEmerald:
Huh, how does that influence the stop time?

100GPing100:
`LongDamping` for forward/backward movement and `LatDamping` for sideways movement.

HellDragon-HK:
Doesn't that slow down it's overall speed too though by slowing it's acceleration?

100GPing100:
I didn't feel any noticeable difference, neither here nor the Viper. The viper used to have a really high `LongDamping` (a value of `30`maybe) and it was still fast. From what I understand it should only take effect if we're not thrusting, but I may be mistaken. I'll try an insanely huge `LongDamping` value and come back with the results.

HellDragon-HK:
With the Raptor & Cicada the LongDamping seemed to impact starting to move forward but I could've been wrong

100GPing100:
So changed to the most insane value that wouldn't make me die instantly or teleport everywhere (`2.0`) and I do not see any difference in the top movement speed, I can still get across `ONS-IslandHop` very quickly.

HellDragon-HK:
So it doesn't affect when you first start moving though and how long it takes to reach max speed?

100GPing100:
Nope, just tested a value of `2.0` for both properties against a value of `0.1`, also for both, and it took the same time from the same starting point to the same end point (aka same time since it starts moving to reach the same end point in a straight line).

HellDragon-HK:
Hm guess it functions differently for Raptor & Cicada then

GreatEmerald:
Huh, good to know that this is the variable that affects it.

Unfortunately looks like this needs a rebase anyway as well.

100GPing100:
Hmm, this doesn't seem right? @GreatEmerald did I just fuck it up? xD

100GPing100:
Okay, here's what I'm gonna do. I'm going to delete my branch and create a new one, do the changes, commits, create new pull request. Right now this just got messy.

PR #161, opened 2017-12-31
Update UT3PaladinCannon.uc

HellDragon-HK:
Override stock value of 0.8 which was making the turret smaller than it should be at 1.35


GreatEmerald:
Nice catch, thanks.

Issue #125, opened 2017-12-27
Nightshade Turret


HellDragon-HK:
- SOUND RELATED
  - [ ] Impact sound of beam weapon is bullet impact sounds in UT2004, doesn't seem to really have an impact sound in UT3 that I can hear

 - VISUAL EFFECTS RELATED
  - [ ] Becomes very bright and easy to see in UT3 when firing beam, doesn't in UT2004

100GPing100:
At the moment I can't fix `Becomes very bright and easy to see in UT3 when firing beam, doesn't in UT2004` a material needs to be created, something I can't currently do on Linux. Maybe someday if I spend some time on this on my Windows machine.

Issue #10, opened 2017-11-27
Nightshade


HellDragon-HK:
Issues and notes on the Nightshade from in-game testing

UT2004 Draw  - NA
UT3    Draw  - 1.0?
UT3Port Draw - 1.0 / 1.0 looks correct

Replaces Axon Goliath (On WAR-Downtown_Necris at least)
Has no headlights but neither does the one in UT3 from what I can tell
Other than some minor issues this one feels more complete & ready for prime time

  - [Priority]  
  - [Turret has no collision]  
  - [Pressing jump button enters deploy mode which does work but doesn't seem you can aim like it is in UT3 (side to side)] 
  - [Secondary]  
  - [The front grill like area has team lights as well as the side grills]  
  - [Has kind of flashing team lights on every spike in UT3, nothing in UT2004]  
  - [Becomes very bright and easy to see in UT3 when firing beam, doesn't in UT2004]  
  - [While not invisible on UT3 it has Manta like dust blow under it, not in UT2004 though]  
  - [Extra]

Unre-Alex:
Nightshade doesn't have functioning stasis field.

HellDragon-HK:
Ah yeah forgot I meant to list those, if I remember right the deployables had a Coming Soon in Next Release message so wasn't sure it was needed to be listed

GreatEmerald:
Hah, that message is by @100GPing100, but it kind of depends on someone coding the thing. The deployables should probably be their own issue though.

HellDragon-HK:
I was going to test and add each one to the Nightshade or Stealthbender list when I got to those vehicles (except I can't use the Stealthbender at this time) but if you want it separate I'll do that unless someone beats me to it

100GPing100:
@GreatEmerald @HellDragon-HK You guys talking about deployables as in weapon like? If so, the actual deploy is already done, would be a matter of adding a new "weapon" that would pawn them on fire. Thing is the deployables in UT3 do not let you change to other weapons, I don't know how trivial this would be to implement in UT2004.

In regards to the stasis field as mentioned by @Unre-Alex I did not find a way for this to be implemented at the time. The way it works in UT3 is that basically everything has a property that is their game speed modifier and the engine uses this to calculate their delta time, in UT2004 we do not have anything like this (at least I did not find anything at the time) and I could not think of a way to implement it.

HellDragon-HK:
[...]

PR #163, opened 2018-01-01
Nightshade patch

100GPing100:
Here's a list of changes:
- Can no longer change mine while deploying/undeploying
- Forced and new third person view while deployed
- Arm rotation while deployed
- State changing code improved
- Beam damage type (fixes the bullet hit sound bug)
- Added dust/water effects while uncloaked
- Renamed `Weap_UT3Nightshade` to `UT3Weap_NightshadeBeam`
- Beam gun no longer rotates while deployed

Fixed from #10:
- [X] ~~Pressing jump button enters deploy mode which does work but doesn't seem you can aim like it is in UT3 (side to side)~~
- [X] ~~While not invisible on UT3 it has Manta like dust blow under it, not in UT2004 though~~

Fixed from #125:
- [X] ~~Impact sound of beam weapon is bullet impact sounds in UT2004, doesn't seem to really have an impact sound in UT3 that I can hear~~

### DIFF
--- a/Classes/UT3DmgType_NightshadeBeam.uc
+++ b/Classes/UT3DmgType_NightshadeBeam.uc
@@ -1,6 +1,5 @@
 /*
- * Copyright © 2012 100GPing100
- * Copyright © 2014 GreatEmerald
+ * Copyright © 2017 Luís 'zeluisping' Guimarães (100GPing100)
  *
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions are
@@ -47,7 +46,7 @@ DefaultProperties
 	MaleSuicide="%o shafted himself.";
 	FemaleSuicide="%o shafted herself.";
 
-	WeaponClass=class'Weap_UT3Nightshade';
+	WeaponClass=class'UT3Weap_NightshadeBeam';
 
 	bDetonatesGoop=true
     bCausesBlood=false
--- a/Classes/UT3DmgType_NightshadeBeam.uc
+++ b/Classes/UT3DmgType_NightshadeBeam.uc
@@ -1,0 +1,67 @@
+/*
+ * Copyright © 2012 100GPing100
+ * Copyright © 2014 GreatEmerald
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are
+ * met:
+ *
+ *     (1) Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *
+ *     (2) Redistributions in binary form must reproduce the above copyright
+ *     notice, this list of conditions and the following disclaimers in
+ *     the documentation and/or other materials provided with the
+ *     distribution.
+ *
+ *     (3) The name of the author may not be used to
+ *     endorse or promote products derived from this software without
+ *     specific prior written permission.
+ *
+ *     (4) The use, modification and redistribution of this software must
+ *     be made in compliance with the additional terms and restrictions
+ *     provided by the Unreal Tournament 2004 End User License Agreement.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE AUTHOR ``AS IS'' AND ANY EXPRESS OR
+ * IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR ANY DIRECT,
+ * INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+ * STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING
+ * IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ *
+ * This software is not supported by Atari, S.A., Epic Games, Inc. or any
+ * of such parties' affiliates and subsidiaries.
+ */
+
+class UT3DmgType_NightshadeBeam extends WeaponDamageType
+	abstract;
+
+DefaultProperties
+{
+	DeathString="%o was carved up by %k's Nightshade shaft.";
+	MaleSuicide="%o shafted himself.";
+	FemaleSuicide="%o shafted herself.";
+
+	WeaponClass=class'Weap_UT3Nightshade';
+
+	bDetonatesGoop=true
+    bCausesBlood=false
+	bLeaveBodyEffect=true
+	bSkeletize=true
+
+    //DamageOverlayMaterial=Material'XGameShaders.PlayerShaders.LinkHit'
+    //DamageOverlayTime=0.5
+
+	//DeathOverlayMaterial=Material'XGameShaders.PlayerShaders.LinkHit'
+    //DeathOverlayTime=1.0
+
+	VehicleMomentumScaling=0.1;
+	VehicleDamageScaling=0.8;
+
+	KDamageImpulse=100
+}
--- a/Classes/UT3Nightshade.uc
+++ b/Classes/UT3Nightshade.uc
@@ -283,7 +283,10 @@ function Tick(float DeltaTime)
 	CheckState();
 
 	if (CurrentState == VS_Deployed && PlayerController(Controller) != none) {
-		ShoulderRotation.Yaw = (Rotation - Controller.Rotation).Yaw % 65536; // make it [0;65536]
+		ShoulderRotation.Yaw = (Rotation - Controller.Rotation).Yaw % 65536;
+		if (ShoulderRotation.Yaw < 0) {
+			ShoulderRotation.Yaw += 65536;
+		}
 
 		if (ShoulderRotation.Yaw <= 32768 && ShoulderRotation.Yaw > 8192) {
 			ShoulderRotation.Yaw = 8192;
--- a/Classes/UT3Nightshade.uc
+++ b/Classes/UT3Nightshade.uc
@@ -1,6 +1,6 @@
 /*
- * Copyright © 2012 100GPing100
- * Copyright © 2014 GreatEmerald
+ * Copyright ï¿½ 2012 100GPing100
+ * Copyright ï¿½ 2014 GreatEmerald
  *
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions are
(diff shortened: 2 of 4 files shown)


Unre-Alex:
The dust effect stays after you exit the vehicle, and it should be a little faster, especially when cloaked. Is there a reason why it doesn't rotate its arm 90 degrees?

Classes/UT3DmgType_NightshadeBeam.uc
```diff
@@ -0,0 +1,66 @@
...
+DefaultProperties
+{
+	DeathString="%o was carved up by %k's Nightshade shaft.";
+	MaleSuicide="%o shafted himself.";
+	FemaleSuicide="%o shafted herself.";
+
+	WeaponClass=class'UT3Weap_NightshadeBeam';
+
+	bDetonatesGoop=true
+    bCausesBlood=false
```
GreatEmerald: These need tabs to spaces.

Classes/UT3Nightshade.uc
```diff
@@ -1,6 +1,6 @@
 /*
- * Copyright � 2012 100GPing100
- * Copyright � 2014 GreatEmerald
+ * Copyright � 2012, 2017, 2018 Luís 'zeluisping' Guimarães (100GPing100)
+ * Copyright � 2014 GreatEmerald
```
GreatEmerald: The copyright symbols got messed up. Make sure all files are in UTF-8 encoding.

Classes/UT3Nightshade.uc
```diff
@@ -91,8 +92,39 @@ var float LastDropAttemptTime;
...
+/* The third person camera distance before we deployed. */
+var float LastDesiredTPCamDistance;
+/* The third person camera distance to use when deployed. */
+var float DeployedTPCamDistance;
+/* The third person camera position to use when deployed. */
+var vector DeployedTPCamLookat;
 
 
+
+simulated function Destroyed() {
```
GreatEmerald: I try to keep the style consistent with UT2004, i.e. every `{` should be on its own line and indents are four spaces rather than a tab.

Classes/UT3Nightshade.uc
```diff
@@ -91,8 +92,39 @@ var float LastDropAttemptTime;
...
+	super.Destroyed();
+
+	if (ArmMine != none) {
+		ArmMine.Destroy();
+	}
```
GreatEmerald: Braces for single lines are not needed.

Classes/UT3Nightshade.uc
```diff
@@ -156,7 +174,7 @@ function bool NoObstacle()
 	else
 		return false;
 }
-/* � Type:
+/* � Type:
```
GreatEmerald: This character got messed up too.

Classes/UT3Nightshade.uc
```diff
@@ -255,21 +267,37 @@ function Fire(optional float F)
...
+	if (CurrentState == VS_Deployed && PlayerController(Controller) != none) {
+		ShoulderRotation.Yaw = (Rotation - Controller.Rotation).Yaw % 65536;
+		if (ShoulderRotation.Yaw < 0) {
+			ShoulderRotation.Yaw += 65536;
+		}
+
+		if (ShoulderRotation.Yaw <= 32768 && ShoulderRotation.Yaw > 8192) {
+			ShoulderRotation.Yaw = 8192;
+		} else if (ShoulderRotation.Yaw > 32768 && ShoulderRotation.Yaw < 57344) {
+			ShoulderRotation.Yaw = 57344;
```
GreatEmerald: It feels like the `Clamp` functions should be able to handle this with less text.

Classes/UT3Nightshade.uc
```diff
@@ -278,91 +306,160 @@ function CheckState()
...
+			break;
+		case VS_Undeploying:
+			SetBoneRotation('ArmShoulder', rot(0,0,0), 0, 1);
+			PlaySound(UndeploySnd, SLOT_None);
+			PlayAnim('ArmRetract', 4.8, 0.2);
+			CurrentAnim = "ArmRetract";
+			break;
+		default:
+			// crash
+			break;
```
GreatEmerald: There are no other states, so this will never be run. Instead, you should make `VS_Undeployed` and `VS_Cloaked` the default, that will save some lines of code.

Classes/UT3Nightshade.uc
```diff
@@ -278,91 +306,160 @@ function CheckState()
...
 	}
 }
-function Cloak(bool OnOff)
-{
+function Cloak(bool OnOff) {
```
GreatEmerald: Booleans should be prefixed with `b` and always be positive. In this case it would make sense to call it `bCloak`.

Classes/UT3Nightshade.uc
```diff
@@ -278,91 +306,160 @@ function CheckState()
...
-		}
-		Visibility = Default.Visibility;
-		bDrawVehicleShadow = true;
-		KarmaParamsRBFull(KParams).KMaxSpeed = MaxVisibleSpeed;
-	}
-	else
-	{
+	local int i;
+
+	if (OnOff == true) {
```
GreatEmerald: Using `== true` is useless and just wastes a CPU cycle, simply remove that part.

Classes/UT3Weap_NightshadeBeam.uc
```diff
@@ -1,6 +1,6 @@
 /*
- * Copyright � 2012 100GPing100
- * Copyright � 2014 GreatEmerald
+ * Copyright � 2012, 2017, 2018 Luís 'zeluisping' Guimarães (100GPing100)
+ * Copyright � 2014 GreatEmerald
```
GreatEmerald: Same here, check the encoding

100GPing100:
@Unre-Alex The angle the arm rotates at is clamped to `45º`, I used this value from memory thinking it was what happens in UT3, so you say it's `90º`? Confirm this for me please and I'll fix it.

Thank you for spotting the dust effects not being destroyed when leaving the vehicle, completely forgot to take care of that!

Regarding the movement speed, the speed while cloaked is a percentage of the speed when not cloaked (just like in UT3 and using the same percentage) so what we can change is the uncloaked speed which will also affect the cloaked speed. Unfortunately I can't fix this one without getting values from you guys, maybe a discussion on #10 or even a different pull request?

Classes/UT3DmgType_NightshadeBeam.uc
100GPing100: God damnit, need to check my editor settings!

Classes/UT3Nightshade.uc
100GPing100: This is weird because that's exactly how it shows in my editor and encoding is set to UTF-8, I even changed to it manually. Gotta figure out what's happening here, thanks for pointing this out!

Classes/UT3Nightshade.uc
100GPing100: Sorry on this one, habit kicked in on this one!

Classes/UT3Nightshade.uc
100GPing100: Another habit, will change! Think I may have some more that can be removed

Classes/UT3Nightshade.uc
100GPing100: That's weird, didn't even know this character was special

Classes/UT3Nightshade.uc
100GPing100: The thing here is that the clamp value depends on which half of the unit circle the angle is on:
image
The only other way I could think of would be to add `8192` (`45º`) to the angle, this would give me a clamp range from `0` to `16384` (`0º` to `90º`), then after clamping I'd remove the added `8192` (`45º`). For clarity I left the if approach, but I'd gladly do it this way, in fact I prefer it way more and I can leave a comment of some sort to explain it.

HellDragon-HK:
Haven't tried this yet or even really touched the Necris vehicles but the arm is able to move this far in UT3 100GP if you need visual confirmation which can move by mouse or keyboard
screenshot00000
screenshot00001ns

Classes/UT3Nightshade.uc
100GPing100: Will do!

GreatEmerald:
The speed values are here:
http://greatemerald.xmpcommunity.com/uncodex/UT3-15/Source_utgame/utvehicle_nightshade.html#62

We've determined that they should match between UT3 and UT2004.

Classes/UT3Nightshade.uc
100GPing100: Okay, but now I feel stupid for not having renamed this variable xD

Classes/UT3Nightshade.uc
100GPing100: Again, force of habit, will change!

100GPing100:
@HellDragon-HK You can move it with the keyboard? How exactly? :open_mouth:  Maybe this can be implemented too (together with the camera constraint that should be implemented at some point). Thank you for the visual confirmations, but it's not really necessary, I trust your word, just need to know if it's confirmed! Thank you!

HellDragon-HK:
That's odd I distinctly remember Forward and Backward keys moving it and it doesn't now, I'm positive I did it it that way because mouse movement is harder and not as quick but I haven't played since right after the Titan Pack came out and there are a lot of oddities since the Game Spy patch I'm noticing I don't remember

Classes/UT3Nightshade.uc
GreatEmerald: Yea, that would work. Though now we know that the clamping needs to happen at 16k and 48k.

100GPing100: Yes, I'll be changing that too!

Unre-Alex:
@100GPing100 Well, as you can see HellDragon confirmed that the arm rotates 90 degrees, I thought there was some issue you couldn't make it work. In UT3, you can rotate it with a mouse or 'turn left' and 'turn right' keys (not strafe keys), which works now in UT2004.

The speed values match UT3? I feel like it's noticeably slower in UT2004. Do you want me to mess with the speed and tell you what I feel is right or do we stay with the current values?

GreatEmerald:
Unless you can find a good enough explanation for why the values should be different, they need to stay as in UT3. They might be modified by the code somewhere, but that means the code needs to be changed to do the same modifications, not the values.

Unre-Alex:
@100GPing100 Two more issues. You can't switch between mines when undeployed and there shouldn't be a shadow when cloaked.

100GPing100:
@Unre-Alex 
> You can't switch between mines when undeployed
This means we should be able to select mines while undeployed (cloaked and not cloaked but not while deploying/undeploying) or it means we currently can and shouldn't?

@GreatEmerald @HellDragon-HK Since this is a hover vehicle, HellDragon is right and the feeling @Unre-Alex has is valid; most likely what needs to be changed here is the thrust, `GroundSpeed`, `AirSpeed` and the `KarmaParams`'s `KMaxSpeed` should limit the maximum speed properly.

Unre-Alex:
@100GPing100 

You don't have UT3 installed right?

> This means we should be able to select mines while undeployed (cloaked and not cloaked

Yes, it worked before this patch.

> but not while deploying/undeploying)

In UT3 you can select mines while undeploying, and you only can't while deploying.

Also,

The vehicle shouldn't be able to deploy while moving. In UT3 you got the massage "Cannot deploy while moving.", you can deploy while moving when cloaked tho (you have the deploy icon all the time).

In UT3 you can deploy while on water (even deep water that would cover the vehicle), it only doesn't let you release the slow field.

100GPing100:
@Unre-Alex Yeah, no UT3 installed because the machine I work on is Linux (and if there is UT3 for linux the specs on this machine aren't enough for that).

That's a lot of useful info! Will take care of applying all that, thank you!

GreatEmerald:
Indeed, UT3 was never released on Linux, so it would only work through Wine.

PR #164, opened 2018-01-02
Discussion: Updating Binaries.csv

HellDragon-HK:


GreatEmerald:
Yes, that's about the idea. However, the whole point of this is to make sure that binaries and code never go out of sync, so updates to `Binaries.csv` file should almost always be done as part of the pull request that requires changes to said binaries.

In addition, there are two more things that you should keep in mind for Google Drive downloads:
- The URL you get from it does not allow automatic downloading. Check the very first line: the URL needs to be in the form of `https://drive.google.com/uc?export=download&id=<id>`, whereas you obtain URLs in the form of `https://drive.google.com/open?id=<id>`.
- In the `sha256` field you need to put the `sha256` checksum of each file. This uniquely identifies the file so that the script knows which files need to be replaced and which don't. On Windows it seems easiest to install this: http://implbits.com/products/hashtab/ Then you can go to file properties, `File Hashes`, and copy the `SHA-256` hash (check to make sure that you get the same hash as I have put for the Scorpion anims).

Also, as a matter of policy, please keep older files around on your Google Drive for a bit, so that in case something goes wrong we can go back to the older version. The file names and folders you use on Google Drive don't matter at all, so you can use that to keep several versions of the same files.

HellDragon-HK:
For Google Drive i'm using Get Shareable Link, I don't know any other way of getting the URL

GreatEmerald:
Yes, that's the correct way. You'll just need to manually replace `open?` from the link with `uc?export=download&` (or you can use search and replace). You can also test it: if you open the `open` format link it gives you an overview page, but if you open the `uc` format link it gives you the file straight away.

HellDragon-HK:
I need a step by step guide for doing this please to not make a mistake, I did install the program already

GreatEmerald:
Here, in the Binaries.csv file. Just look at the Files changed tab, at the difference between the line you removed and the topmost line you added. There should be no changes between the two.

HellDragon-HK:
The only thing changed was new lines added and carefully changing the URL so still ?

GreatEmerald:
Yes, that's the whole point. You have put the URL:
https://drive.google.com/open?id=1QSeKP6KxkT9vBNyptGE7ouPl76W1EL-C
It should be:
https://drive.google.com/uc?export=download&id=1QSeKP6KxkT9vBNyptGE7ouPl76W1EL-C

HellDragon-HK:
Ok but what about the sha256 that's the part I don't have any understanding with

GreatEmerald:
As I said above, install HashTab, go to the properties of the file you uploaded to Google Drive, copy the SHA-256 hash, and that's what you add into the `sha256` field.

GreatEmerald:
As in, the properties of that file locally. You can take the hash first, then upload it, then update the `Binaries.csv` file.

HellDragon-HK:
Ok so it comes from my PC files then that's what I was trying to ask

GreatEmerald:
Yeap.

HellDragon-HK:
I don't have an sha256, only CRC32, MD5, SHA-1 and none of those are b66bc22d1df53ae1335c21f5053c08fcb2af7c4554f829a64b3a2ea8618f6b2c

GreatEmerald:
Press the Options button, you should have the ability to choose which hashes to show in the list.

100GPing100:
@HellDragon-HK This value is kinda like a representation of the file. If you get two files that are exactly the same and compare their key they will be the same (just like how the files are the same); so if you have two different files, even just a byte, the key will be different.

So this can be used to know if a file changed. If you have two files (in this case the local file and the downloaded file) and they have different keys, then this means that they are not the same, and since the downloaded file is the one we want to have, we replace the local file with the downloaded one.

So in the end this key is a representation of the file just to be able to compare the two files and know if they are the same or not.

HellDragon-HK:
Ok this looks the same but I don't do well with spotting differences in a mess like this

B66BC22D1DF53AE1335C21F5053C08FCB2AF7C4554F829A64B3A2EA8618F6B2C

GreatEmerald:
And that's why you have the `Hash Comparison` box: put in `b66bc22d1df53ae1335c21f5053c08fcb2af7c4554f829a64b3a2ea8618f6b2c` and it should do the comparing for you! And give a nice green check mark.

HellDragon-HK:
Ok then I'll start editing the request here to put those in

GreatEmerald:
Yeap. In the mean while I'm working on adding the rest of the files, currently done with `System`. I see you have all the animations except for `EONSLocustA` (ideally this would be replaced by the actual hoverboard animations), `UT3NemesisAnims` and `UT3NightshadeAnims`; @100GPing100: you're working on those, don't forget to upload them somewhere and put the links in here too.

HellDragon-HK:
We can use the same G-Drive folder for everything, I just need to give editing access to a specific account

GreatEmerald:
I have my own NextCloud instance, so I don't need that.

HellDragon-HK:
What about Alex's texture work?

He has been using Mediafire for his texture work so far which I don't think is a good idea considering mediafire isn't all that trustworthy (plus MBAM hates it most of the time)

GreatEmerald:
Indeed, we can't use Mediafire here.

100GPing100:
@GreatEmerald I haven't done any binary changes, do you still want me to take care of the nightshade binaries?

HellDragon-HK:
Oh btw thanks for the info @100GPing100

GreatEmerald:
@100GPing100 Good to know, then we'll handle it.

HellDragon-HK:
@100GPing100 I thought you said you fixed the sound issue on the Cicada and were going to do a pull request, didn't that require adding in the missing sound to the uax?

GreatEmerald:
That's fine. Once you edit the files, just upload them again and edit the `Binaries.csv` file in a pull request.

HellDragon-HK:
@GreatEmerald I have Alex's textures installed so would it do any good to set them up (without the link) in the csv?

GreatEmerald:
Not yet. I got the textures covered now, for his textures you'll need to submit a pull request later on so that they can be tested on 64-bit some more.

HellDragon-HK:
Ok but shouldn't that be up to @Unre-Alex , it's his work after all?

GreatEmerald:
Sure, he can do that too.

HellDragon-HK:
I don't mind doing it so it's up to you @Unre-Alex ,I can do it or do you want to?

GreatEmerald:
@HellDragon-HK I put all the files from my side, can you rebase this and sort your own entries so that they are all in alphabetical order?

HellDragon-HK:
What does rebase mean?

GreatEmerald:
I updated the file, and you updated the old version of the file. The pull request can't be merged until your changes are to the new version of the file. Rebasing is updating your changes to be applicable to the new version.
You can also use the "Resolve conflicts" button down here to do that.

HellDragon-HK:
Ok done, noticed --content-disposition is missing am I supposed to remove that too?

GreatEmerald:
No, `--content-disposition` is needed for Google Drive, not for other sources.

HellDragon-HK:
Ok then so anything I do it 'has' to be there, good to know as you say

GreatEmerald:
Hm, the rebase looks OK but somehow GitHub still shows that there are conflicts? Does it show the same here for you too, or is it something on my end?

HellDragon-HK:
The conflict resolve never works for me so yes it's still grey, nothing to do but start over right?

GreatEmerald:
No no, your fork is fine, it's just GitHub being silly. I can fix it by just opening a new pull request.

HellDragon-HK:
Alright I'll let you handle it, btw this is the issue I had with your edit on the Raptor plasma, it really didn't show as edited, GitHub has always been noticeably slow lately

GreatEmerald:
I think I need to close this request to open a new one. Don't delete the branch in the mean while.

100GPing100:
Sorry for comment on closed issue:

> @100GPing100 I thought you said you fixed the sound issue on the Cicada and were going to do a pull request, didn't that require adding in the missing sound to the uax?
I haven't touched the Cicada yet, plus creating a package would be troublesome on my end, all good on this one.

HellDragon-HK:
Ah sorry then I must have misunderstood

PR #166, opened 2018-01-02
Hellbender: Fix Sky Mine Combo

HellDragon-HK:


GreatEmerald:
Looks OK, but needs to have the full license to be attached to the top like in other files.

Also, I'd suggest not opening more pull requests until all binaries are purged or else you will have to redo everything over again.

HellDragon-HK:
'Looks OK, but needs to have the full license to be attached to the top like in other files.'   ???

'Also, I'd suggest not opening more pull requests until all binaries are purged or else you will have to redo everything over again.'

I thought that everything now was right and everything then was needing to be started over?

GreatEmerald:
No, it's the other way round.

The license must be there in every file and it's not there right now. It's the whole thing as listed in LICENSE.TXT, except for the copyrights.

And anything not merged before the binary purge will have to be remade. So focus on getting these pull requests merged first.

GreatEmerald:
OK, thanks.

PR #167, opened 2018-01-02
Manta: Engine Pitch & Stop Time

100GPing100:
Some cleanup, engine pitch reduced and stop time fixed.

Supersedes #160 

Fixes from #9:
- [x] ~~UT3 Manta stops on a dime without the need to move in the opposite direction, UT2004 Manta takes a little longer~~
- [x] ~~Movement sound in UT2004 is too high pitched from UT3~~

and maybe, unconfirmed:
- Manta hum in UT3 is actually very quiet and calm when sitting still and in UT2004 is very loud and quite possibly not even the same sound

### DIFF
--- a/Classes/UT3Manta.uc
+++ b/Classes/UT3Manta.uc
@@ -292,6 +292,7 @@ defaultproperties
         
     MaxStrafeForce=27 //20 def UT2004
     LatDamping=0.2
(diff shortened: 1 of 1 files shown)


GreatEmerald:
OK, thanks.

PR #168, opened 2018-01-03
Manta: Fix typo

100GPing100:
This fixes the typo from the last manta patch, as was pointed out by @Unre-Alex on commit c821c3e.


GreatEmerald:
Oh man, hotfixes :D

PR #169, opened 2018-01-04
Paladin: Shield Scale Re-work

HellDragon-HK:


GreatEmerald:
So how does it look with `(X=-50,Z=-20)`?

And I still don't get what the `(X=400,Z=140)` values in UT3 could be... But yea, looking at the screenshots, it looks like the shield is supposed to be about 1 Paladin away from the cannon.

HellDragon-HK:
Hah 1 Paladin away, that's funny....anyway here is what -50 on X looks like Z is already -20....seems a little close to me but it could work maybe.

shot00637
shot00636
shot00638

GreatEmerald:
I'm comparing these to what you have in #85. Can you get the angle similar to what you had in https://user-images.githubusercontent.com/34013048/34458643-ef6492fc-ed9f-11e7-8bf1-0a9e474941a9.jpg (or, conversely, take a screenshot in UT3 with something close to what you posted above)? It seems to be close, but since in the other screenshots it's looking the other way and the space is about that of 1 Paladin, and in UT2004 it goes slightly further than the front of it when aiming to the front, it might actually be the right distance.

HellDragon-HK:
I can't get the same front shot in UT3, the shield deactivates if it starts to go under the ground in UT3 but I can try for a similar shot in UT2004

HellDragon-HK:
Ok here is with -50

shot00641

PR #170, opened 2018-01-04
Scorpion: Overall Improvements Re-work

HellDragon-HK:
Ok I think I've got everything back in here, also something I noticed in UT3 is that eject doesn't actually eject, you just spawn in the air and fall but I like how it's currently done in UT2004 which looks way better in my opinion


GreatEmerald:
OK, thanks.

PR #171, opened 2018-01-04
Hellfire SPMA: Various Updates

HellDragon-HK:
Found in the UT3 Editor's SPMA sounds, this is either very quiet in-game or unused


GreatEmerald:
Hm, I'm not seeing this in the code. There is a sound for rotating the main cannon, however (`A_Vehicle_SPMA.SoundCues.A_Vehicle_SPMA_CannonRotate`), is that the same sound?

HellDragon-HK:
No it's not, this is something that was I guess unused or used to be used.

GreatEmerald:
I'm guessing that it was unused, then. In which case I don't think there's a good reason to reintroduce it if it doesn't really improve anything.

HellDragon-HK:
I think everything is working now including the camera's ambient sound...though possibly a little loud, the engine on the other hand is the opposite and like usual is probably going to be too quiet.

HellDragon-HK:
Argh this Travis thing hates me, I just uploaded that binary and got the checksum from the same file so how did it fail, a glitch?

HellDragon-HK:
Full update of non-recorded varied sounds in Basic or Extended flavor

Note: Haven't done Goliath yet, it's next on the list so when it's done the impact sound here needs to be changed to use that

Also need to look at fixing the no shock combos issue that I had with the Hellbender

HellDragon-HK:
Fixed the No Combos problem

HellDragon-HK:
Added suspensions fix to get SPMA out of the ground and reorganized defaults into an imo much more readable list with clear identity of entries....categorization may need some further adjusting though.....still a trial run so to speak.

PR #173, opened 2018-01-04
Raptor: Plasma Improvements Cont.

HellDragon-HK:
Continuing from where we left off we should be getting 40 damage per shot to vehicles but it only seems to happen on everything shot after the first shot which only seems to do 20.

### DIFF
--- a/Classes/UT3RaptorProjBlue.uc
+++ b/Classes/UT3RaptorProjBlue.uc
@@ -102,24 +102,35 @@ class UT3RaptorProjBlue extends ONSAttackCraftPlasmaProjectileBlue;
     bHurtEntry = false;
 }*/
 
+simulated function UpdateDamage()
+{
+    if ( LifeSpan < 1.0 )
+        Damage = 0.75 * default.Damage;
+    else
+        Damage = default.Damage * FMin(2.0, Square(MaxSpeed)/Square(Speed));
+}
+
+simulated function HitWall(vector HitNormal, actor Wall)
+{
+    UpdateDamage();
+    Super.HitWall(HitNormal, Wall);
+}
+
 simulated function Explode(vector HitLocation, vector HitNormal)
 {
-    local float MyDamage;
-
-    MyDamage = 20.0;
-
-    if ( LifeSpan < 1.0 )
-        Damage = 0.75 * MyDamage;
-    else
-        Damage = MyDamage * FMin(2.0, Square(MaxSpeed)/Square(Speed));
-
-    default.Damage = Damage;
+    UpdateDamage();
     Super.Explode(HitLocation, HitNormal);
+}
+
+simulated function BlowUp(vector HitLocation)
+{
+    UpdateDamage();
+    Super.BlowUp(HitLocation);
 }
 
 defaultproperties
 {
-    Damage = 20.0 //GE: We're weaker! Hooray!
+    Damage = 20.0 //20 GE: We're weaker! Hooray!
     Speed=2000
     MaxSpeed=12500
     AccelerationMagnitude = 20000 //GE: And slower!
--- a/Classes/UT3RaptorProjBlue.uc
+++ b/Classes/UT3RaptorProjBlue.uc
@@ -1,5 +1,6 @@
 /*
  * Copyright © 2009, 2014 GreatEmerald
+ * Copyright © 2017-2018 HellDragon
  *
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions are
@@ -101,10 +102,27 @@ class UT3RaptorProjBlue extends ONSAttackCraftPlasmaProjectileBlue;
     bHurtEntry = false;
 }*/
 
+simulated function Explode(vector HitLocation, vector HitNormal)
+{
+    local float MyDamage;
+
+    MyDamage = 20.0;
+
+    if ( LifeSpan < 1.0 )
+        Damage = 0.75 * MyDamage;
+    else
+        Damage = MyDamage * FMin(2.0, Square(MaxSpeed)/Square(Speed));
+
+    default.Damage = Damage;
+    Super.Explode(HitLocation, HitNormal);
+}
+
 defaultproperties
 {
     Damage = 20.0 //GE: We're weaker! Hooray!
+    Speed=2000
+    MaxSpeed=12500
     AccelerationMagnitude = 20000 //GE: And slower!
-    MomentumTransfer = 4000.0
+    MomentumTransfer = 20000.000000 //4000.0
     LifeSpan = 1.6
 }
--- a/Classes/UT3RaptorProjRed.uc
+++ b/Classes/UT3RaptorProjRed.uc
@@ -1,5 +1,6 @@
 /*
  * Copyright © 2009, 2014 GreatEmerald
+ * Copyright © 2017-2018 HellDragon
  *
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions are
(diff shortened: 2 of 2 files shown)


GreatEmerald:
And if you change `Damage` in defaultproperties to, say, 200, does the first shot always deal 200?

HellDragon-HK:
I guess we can find out

HellDragon-HK:
Considering the Goliath has 900 and it had 700 when I got in I'd say yes

GreatEmerald:
Not really, I'm trying to find out what's going on here. This still doesn't seem to make sense, even though there's a method to the madness. It looks like the first shot deals `default.Damage` no matter anything else, as if it's not even going through the `Explode` function or so, because it would otherwise be overridden by the `default.Damage` setting in the function...

If you remove the `default.Damage = Damage;` line, does the behaviour change any? You can keep the damage defaultproperties at 200 to make it obvious.

HellDragon-HK:
Removing that now has consistent 200 damage, Manta destroyed in single shot, Raptor set on fire, Scorpion set on fire, Goliath dropped to 700 hp and Hellbender dropped to 400 hp

GreatEmerald:
And if you remove the `Super.Explode()`, does it deal no damage always?

HellDragon-HK:
The whole line or just Super.Exlode and leave (HotLocation, HitNormal); ?

HellDragon-HK:
'does it deal no damage always'

Uh actually the opposite, a Goliath and every vehicle after is exploding in a single shot

GreatEmerald:
That's also weird, maybe everything dies in one shot because the game thinks it's a telefrag... Put `Destroy();` at the end of the function and see whether it does any damage then.

HellDragon-HK:
With or without Super.Explode?

GreatEmerald:
Without. It should not deal any damage...

HellDragon-HK:
simulated function Explode(vector HitLocation, vector HitNormal)
{
    local float MyDamage;

    MyDamage = 20.0;

    if ( LifeSpan < 1.0 )
        Damage = 0.75 * MyDamage;
    else
        Damage = MyDamage * FMin(2.0, Square(MaxSpeed)/Square(Speed));

    //default.Damage = Damage;
    Destroy();
}

defaultproperties
{
    Damage = 200.0 //20 GE: We're weaker! Hooray!
    Speed=2000
    MaxSpeed=12500
    AccelerationMagnitude = 20000 //GE: And slower!
    MomentumTransfer = 20000.000000 //4000.0
    LifeSpan = 1.6
}

here you go, this is what I have right now and it's causing a consistent 200 on everything

HellDragon-HK:
Wait, how did that get back to 200 when I just set it at 20!?

GreatEmerald:
Interesting, I think the reason this is happening is because we're looking at the wrong function entirely! Try this:
```UnrealScript
simulated singular function HitWall(vector HitNormal, actor Wall)
{
    Damage = 123;
    Super.HitWall(HitNormal, Wall);
}
```
This ought to deal 123 damage to vehicles.

HellDragon-HK:
Remove the Explode function?

GreatEmerald:
No, keep it for the time being.

HellDragon-HK:
Ok kind of makes things hard as there is no impact visual now

HellDragon-HK:
Well now you've got the 0 damage happening, also not being told I can't hurt unlinked core

GreatEmerald:
Argh, it must be due to the `singular`ity... Try this instead:
```UnrealScript
simulated function HitWall(vector HitNormal, actor Wall)
{
    Damage = 123;
    Super.HitWall(HitNormal, Wall);
}
```

HellDragon-HK:
Ok that worked, my math may be off but I think it's doing 123 now, all vehicles have _77 after a shot, core shield still isn't telling me I can't hurt it and still can't visually tell that I'm hitting anything unless it catches on fire or starts smoking

GreatEmerald:
Awesome. So then it should all look like this, I suppose:

```UnrealScript
simulated function UpdateDamage()
{
    if ( LifeSpan < 1.0 )
        Damage = 0.75 * default.Damage;
    else
        Damage = default.Damage * FMin(2.0, Square(MaxSpeed)/Square(Speed));
}

simulated function HitWall(vector HitNormal, actor Wall)
{
    UpdateDamage();
    Super.HitWall(HitNormal, Wall);
}

simulated function Explode(vector HitLocation, vector HitNormal)
{
    UpdateDamage();
    Super.Explode(HitLocation, HitNormal);
}

simulated function BlowUp(vector HitLocation)
{
    UpdateDamage();
    Super.BlowUp(HitLocation);
}

```

HellDragon-HK:
Unfortunately nope, 350 nearly on the first shot to a Hellbender and all other vehicles 123 by the look of it

GreatEmerald:
Uh, do make sure you have `Damage` in defaultproperties set back to 20. Otherwise update the pull request with what you have, because that seems implausible.

HellDragon-HK:
I did, it already is set at 20 and has been since I noticed it earlier

GreatEmerald:
So there's no way it would deal 123, there's no place for it to take such a number... Well, place this at the bottom of the `UpdateDamage()` function:
```UnrealScript
log(self@"UpdateDamage: Dealing"@Damage@"instead of"@default.Damage)
```

HellDragon-HK:
Sure but log hasn't helped much so far, what we get in-game including what Alex got isn't what the log says so far....why is that and how is that possible?

GreatEmerald:
It isn't. The reason why it wasn't very useful before is because we've been editing the wrong function, so it wouldn't log anything at all (you saw logs of shots to other things that you didn't check the health of).

HellDragon-HK:
Yo mean the floor was taking 40 damage?

GreatEmerald:
Exactly.

And yes, you need to add `;` at the end.

HellDragon-HK:
Lol I wasn't even aware the floor could take damage.

The heck, now I'm getting 40 on every shot...but the only thing changed was adding a log

ScriptLog: ONS-RedPlanet.UT3RaptorProjRed UpdateDamage: Dealing 15.00 instead of 20.00
ScriptLog: ONS-RedPlanet.UT3RaptorProjRed UpdateDamage: Dealing 15.00 instead of 20.00
ScriptLog: ONS-RedPlanet.UT3RaptorProjRed UpdateDamage: Dealing 15.00 instead of 20.00
ScriptLog: ONS-RedPlanet.UT3RaptorProjRed UpdateDamage: Dealing 15.00 instead of 20.00
ScriptLog: ONS-RedPlanet.UT3RaptorProjRed UpdateDamage: Dealing 40.00 instead of 20.00
ScriptLog: ONS-RedPlanet.UT3RaptorProjRed UpdateDamage: Dealing 40.00 instead of 20.00
ScriptLog: ONS-RedPlanet.UT3RaptorProjRed UpdateDamage: Dealing 40.00 instead of 20.00
ScriptLog: ONS-RedPlanet.UT3RaptorProjRed UpdateDamage: Dealing 40.00 instead of 20.00
ScriptLog: ONS-RedPlanet.UT3RaptorProjRed UpdateDamage: Dealing 40.00 instead of 20.00
ScriptLog: ONS-RedPlanet.UT3RaptorProjRed UpdateDamage: Dealing 40.00 instead of 20.00

HellDragon-HK:
Looks like those changes might be good then and since you already had me add it to the request it's already there.....might be best to wait for @Unre-Alex to try it and see what he gets just in case though?

Unre-Alex:
It works.

@HellDragon-HK  You forgot this for the red

```
simulated function UpdateDamage()
{
    if ( LifeSpan < 1.0 )
        Damage = 0.75 * default.Damage;
    else
        Damage = default.Damage * FMin(2.0, Square(MaxSpeed)/Square(Speed));
}
```

HellDragon-HK:
Thanks, why is it GitHub keeps doing weird things on me, I know for fact I pasted the whole thing....just like how that other class shouldn't have had redundant data when all of that section was replaced

GreatEmerald:
Well, mystery solved, then! Who knew that vehicles were considered to be "walls"...

HellDragon-HK:
I didn't catch that, I caught the floor but not that....hmm maybe it's not so much the vehicles as it is the collision boxes which probably could be considered a wall...just a guess

GreatEmerald:
Yes, probably.

PR #174, opened 2018-01-04
Update UT3Scorpion.uc

HellDragon-HK:
Quick fix for redundant data issue (Thanks Alex, I have no idea how that happened as I pasted over everything about the coronas and I don't have redundant data in my uc file....odd oh well should be fixed with this)


GreatEmerald:
OK.

Sometimes I wonder about setting up a CI to catch these sort of issues. But oh well.

PR #183, opened 2018-01-08
Removed duplicates from the main branch

Unre-Alex:


HellDragon-HK:
What duplicate? But we just said they were missing and re-uploaded them a few hours ago :S

GreatEmerald:
Hm, why is this closed? I can indeed see that the reason why the files were missing is because they were uploaded to the wrong directory. Of course, I can fix this manually just as well.

GreatEmerald:
Oh, superseded by #184. Please indicate that next time ;)

PR #186, opened 2018-01-09
Fix for missing sound references

Unre-Alex:


Unre-Alex:
done, unless you want me to change tabs outside defaultproperties too.

GreatEmerald:
That's fine, thanks.

PR #187, opened 2018-01-13
Manta: Fix Headlight Placement....Again

HellDragon-HK:


GreatEmerald:
Hm, OK. Not sure why it was different before...

HellDragon-HK:
Apparently FlashFlare and FlareOY are the difference as I had to move every headlight on every vehicle after switching to FlashFlare texture.

PR #188, opened 2018-01-13
Manta: Engine Flames WIP, Help Needed

HellDragon-HK:
DO NOT MERGE, I need help solving an issue first

Looking at a mod Manta, I got the trail effects to work but the Manta has reduced jump height now and I can't figure out why, and there is nothing I can see that would cause it although from a learning perspective I'm suspecting it to be an overridden function that is now incomplete.
 
I also am aware the UT3 Manta doesn't have streamers but considering the Manta is a fast vehicle I'd like to keep that for myself at least which I think looks cool and makes sense, I've positioned the streamers to have 2 on the ailerons and 2 below the engines where some small wings are.   

It also looks like I may have found what's needed to fix the Manta's high pitch sound issue but not sure how to work with it.

It's almost as if the Manta is now heavier, even getting in it doesn't jump up as much, it does seem this is the only negative impact so far which if necessary could probably tweak default properties to get something that works.

### DIFF
--- a/Classes/UT3Manta.uc
+++ b/Classes/UT3Manta.uc
@@ -64,6 +64,11 @@ var()	float							StreamerOpacityMax;
 var		float							StreamerCurrentOpacity;
 var		bool							StreamerActive;
 
+var()	bool				bMakeBrakeLights;
+var()	vector				BrakeLightOffset[2];
+var		ONSBrakelightCorona	BrakeLight[2];
+var()	Material			BrakeLightMaterial;
+
 /* The spining blades. */
 var array<UT3MantaBlade> Blades;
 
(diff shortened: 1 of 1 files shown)


HellDragon-HK:
Attempted to add brakelight but won't compile with the error

C:\Unreal Anthology\UT2004\UT3Vehicles\Classes\UT3Manta.uc(288) : Error, Call to 'UpdateBrakelightState': missing or bad parameter 2

Previously before this error was 

C:\Unreal Anthology\UT2004\UT3Vehicles\Classes\UT3Manta.uc(288) : Error, Call to 'UpdateBrakelightState': bad or missing parameter 1

In which after it I tried to change (OutputBrake ,Gear) to (OutputThrust)

GreatEmerald:
It is:
```UnrealScript
simulated function UpdateBrakelightState(float Brake, int Gear)
```
So first is whether it is braking, the second is whether it is going backwards. `Gear` does not exist for hovercraft so you need to figure out how to know whether it is going backwards.

And obviously the streamers cannot be accepted. You might want to make a separate git branch for them and just use it for yourself.

HellDragon-HK:
I was trying to figure out what it was for braking and backwards movement on the Manta when I was working on it's movement....still have no clue what it is...so guess brakelight may be impossible for a Manta.

You didn't say anything about the jumping problem, any idea about that, after that's solved the engine flames might be ready to go.

GreatEmerald:
Hm, I wonder whether `Throttle` might be it. Though it could also be unused.

No, I don't know about the jumping problem. You might want to disable each property/function one at a time to trace that.

HellDragon-HK:
It's the DeltaTime function which what affect does disabling that have, the engine flames still worked as did the blades including turn off and on with and without driver.

GreatEmerald:
Hm, where does the code in `Tick` come from? I'm guessing that you are doing calculations there that are already done in `Super.Tick()` and thus you get twice the effect.

HellDragon-HK:
Tick was already there by GP for the Ailerons....which now that I think of it I don't think work with it disabled, I had to add the extra info to his Tick function because ucc wouldn't accept 2 instances.

HellDragon-HK:
Ok I've noticed the effects of disabling DeltaTime now, the FX are at their max visual and don't ever have their reduced state meaning the engine flames are appearing as they would at full thrust at all times and indeed the ailerons are non-functional.

GreatEmerald:
Hm, the part you added doesn't seem to mess with the physics though. Maybe it's the location of `Super.Tick(DeltaTime)` again and it should be somewhere above?

You should remove the lines that have to do with the pitch, though, otherwise it's calculated twice. Also, you need to fix the indentation in that function, it's pretty hard to read the code right now.

HellDragon-HK:
Moved it to just under the local entries and I think jumping might be higher now but still reduced, might need @Unre-Alex 's opinion on that however.

PR #189, opened 2018-01-13
Hellbender Driver Turret: Corona WIP, Help Needed

HellDragon-HK:
DO NOT MERGE

I've been trying to get a corona for the driver turret but no luck...

C:\Unreal Anthology\UT2004\UT3Vehicles\Classes\UT3HellbenderSideGun.uc(144) : Error, 'bNoTeamBeacon': Bad command or expression

So remove the bNoTeamBeacon = True (which I'm not sure if it's even possible to function without as that's basically if no health bar is true right?) and then I get...

C:\Unreal Anthology\UT2004\UT3Vehicles\Classes\UT3HellbenderSideGun.uc(117) : Error, Unknown Function 'TeamChanged' in 'Class Onslaught.ONSPRVSideGun'

### DIFF
--- a/Classes/UT3HellbenderSideGun.uc
+++ b/Classes/UT3HellbenderSideGun.uc
@@ -39,6 +39,170 @@
 
 class UT3HellbenderSideGun extends ONSPRVSideGun;
 
+#exec obj load file=..\Animations\UT3VH_Hellbender_Anims.ukx
+#exec OBJ LOAD FILE=..\textures\EpicParticles.utx
+#exec OBJ LOAD FILE=..\textures\VMVehicles-TX.utx
+
+var		array<ONSHeadlightCorona>	HeadlightCorona;
+var()	array<vector>				HeadlightCoronaOffset;
+var()	Material					HeadlightCoronaMaterial;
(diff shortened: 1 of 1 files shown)


GreatEmerald:
I'm think the team beacon is the triangle that is shown above vehicles. That part just disables it before destroying the effects.

There is no `TeamChanged` function for a weapon. Instead, you should do these changes on the associated `ONSWeaponPawn` class; which in this case does not exist (no `UT3HellbenderSideGunPawn`, so you should make one. Check how the rear gun pawn works and see the difference between the lines:
https://github.com/GreatEmerald/UT3Vehicles/blob/20113140f4334a117210b8210e1b9e6755742ceb/Classes/UT3Hellbender.uc#L217-L218

HellDragon-HK:
So DriverWeapons will need to become UT3HellbenderSideGunPawn?

GreatEmerald:
Yea, you'll need to replace `WeaponClass` with `WeaponPawnClass`.

HellDragon-HK:
Don't think that;s going to work

Unknown member WeaponPawnClass in DriverWeapons

PR #190, opened 2018-01-13
Manta: Smoke & Fire Locations WIP, Help Needed

HellDragon-HK:
DO NOT MERGE

I've been trying to add the 2nd smoke & fire location as well as set up being able to use a new UT3 visual class later on, so far no luck on getting a 2nd one though.

C:\Unreal Anthology\UT2004\UT3Vehicles\Classes\UT3Manta.uc(67) : Warning, 'DamagedEffect' obscures 'DamagedEffect' defined in base class 'ONSVehicle'.
C:\Unreal Anthology\UT2004\UT3Vehicles\Classes\UT3Manta.uc(68) : Warning, 'DamagedEffectOffset' obscures 'DamagedEffectOffset' defined in base class 'ONSVehicle'.
C:\Unreal Anthology\UT2004\UT3Vehicles\Classes\UT3Manta.uc(69) : Warning, 'DamagedEffectClass' obscures 'DamagedEffectClass' defined in base class 'ONSVehicle'.
C:\Unreal Anthology\UT2004\UT3Vehicles\Classes\UT3Manta.uc(70) : Warning, 'DamagedEffectScale' obscures 'DamagedEffectScale' defined in base class 'ONSVehicle'.

ClassProperty UT3Vehicles.UT3Manta.DamagedEffectClass: unresolved reference to '()'
UT3Vehicles.UT3Manta: Out of bound array default property (1/1)

### DIFF
--- a/Classes/UT3Manta.uc
+++ b/Classes/UT3Manta.uc
@@ -50,6 +50,11 @@ var Emitter DuckEffect;
 #exec obj load file=../Textures/UT3MantaTex.utx
 #exec obj load file=../Sounds/UT3A_Vehicle_Manta.uax
 
+var  	array<ONSDamagedEffect>     DamagedEffect;
+var()	array<vector>               DamagedEffectOffset;
+var()	class<ONSDamagedEffect>     DamagedEffectClass;
+var()	float						DamagedEffectScale;
+
 /* The spining blades. */
 var array<UT3MantaBlade> Blades;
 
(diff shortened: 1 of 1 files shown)


GreatEmerald:
Yea, that won't work. There can only be a single `DamagedEffect`. There are two options: either change the actual particles to show two sources of fire in a single effect, or create something like a `SecondaryDamagedEffect` that would be added and destroyed just like `DamagedEffect` in the `ONSVehicle` class.

HellDragon-HK:
The first option would require a new class for each vehicle wouldn't it?  The 2nd option would then probably be the better choice but how is that done so it actually works right?

GreatEmerald:
A new type of particle emitter, yes. But not for every class, just those that have more than one point.

The second option should just work right if you add the same code as in `ONSVehicle`, just for a differently-named variable. Ideally an array. Hence my suggestion of `SecondaryDamagedEffects`; for each of them, spawn if one does not exist, and destroy on `Destroy`.

HellDragon-HK:
I don't even know where to begin with that, a new UT3ExtraDamagedEffectOne class?  Change 
````var  	array<ONSDamagedEffect>     DamagedEffect;````  to ````var  	array<UT3ExtraDamagedEffectOne>     SecondaryDamagedEffect;```` ?

PR #191, opened 2018-01-13
Manta: Turbine Fire

HellDragon-HK:
Can be merged, at the moment this option works and can be added in as a prelude to having both locations later.


GreatEmerald:
OK, thanks.

PR #193, opened 2018-01-13
Scorpion: Smoke & Fire Locations

HellDragon-HK:
Here are the locations for the smoke and fire effects, only 1 is usable at this time and for this one we need not only 3 but different scales for it to be right.   For now the single body fire point can work as a prelude.


GreatEmerald:
OK, thanks.

PR #194, opened 2018-01-13
Hellbender: Smoke & Fire Locations

HellDragon-HK:
Here are the locations for the Hellbender's smoke & fire, we need 2 places and 2 very different sizes for this one to be right.
For now I think the driver turret location is probably the best option until both can be used.


GreatEmerald:
OK, thanks.

PR #195, opened 2018-01-13
Raptor: Smoke & Fire Locations

HellDragon-HK:
2 Locations, 2 different sizes again, either one of these would be a great single option in my opinion, I think the front fire point turned out really well.


GreatEmerald:
OK, thanks.

PR #197, opened 2018-01-14
Goliath: Engine Flames

HellDragon-HK:
I caught a break on this one and it worked on the first try.   Aside from a Warning this one is ready to go I think, I didn't notice anything odd like how the Manta can't hardly jump now but it probably needs some field testing to be sure, @Unre-Alex @100Gping100 could you guys try it and see how it goes for you?

Here's the warning on compile

C:\Unreal Anthology\UT2004\UT3Vehicles\Classes\UT3Goliath.uc(149) : Warning, 't' : unreferenced local variable

I've also noticed the #exec OBJ Load File has / in all of these classes, but the ones I've seen from other classes have \ , doesn't seem to make any difference to ucc that I can tell so is there a difference?

### DIFF
--- a/Classes/UT3Goliath.uc
+++ b/Classes/UT3Goliath.uc
@@ -46,9 +46,14 @@ class UT3Goliath extends ONSHoverTank;
 
 //=====================
 // @100GPing100
-#exec obj load file=../Animations/UT3GoliathAnims.ukx
-#exec obj load file=../Textures/UT3GoliathTex.utx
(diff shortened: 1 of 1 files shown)


GreatEmerald:
The difference between `\` and `/` might be visible on Linux. Paths there are only `/` and names are all case sensitive (so `Textures`, not `textures`, etc.). This needs testing, then.

HellDragon-HK:
Ok how exactly do I test / and \, I don't think I'm seeing any difference between them with ucc nor in-game.

PR #198, opened 2018-01-14
Hellfire SPMA: Smoke & Fire Locations

HellDragon-HK:
Here are the smoke & fire points for the Hellfire, as usual 2 locations and 2 different sizes are needed for the full effect.


GreatEmerald:
OK, thanks.

PR #200, opened 2018-01-14
Cicada: Smoke & Fire Locations

HellDragon-HK:
Here are the smoke & fire points for the Cicada, 2 locations & 2 sizes again.....both places are good for a prelude and I think I'm going to go with the front fire here.


GreatEmerald:
OK, thanks.

PR #201, opened 2018-01-14
Raptor: Streamer Restoration & Update

HellDragon-HK:
Restoring the streamers was actually a lot easier than I thought, what I was trying to do previously was just completely rebuild them when all that was needed was removing the overridden functions.

The streamers are now back with updated placement and opacity to be closer to UT3.

### DIFF
--- a/Classes/UT3Raptor.uc
+++ b/Classes/UT3Raptor.uc
@@ -219,7 +219,7 @@ function Guns()
 // The next 3 functions { Died, Destroyed, DrivingStatusChanged } have been
 // overriden to disable the streams.
 //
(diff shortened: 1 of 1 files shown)


HellDragon-HK:
What do you mean rebased, this was done after the purge?

GreatEmerald:
I accepted another edit to the same file, hence there's a conflict and you need to rebase this. That happens always when a file is modified in two different ways.

Yea, that's plausible, but I'd like to hear from zeluisping to make sure.

HellDragon-HK:
Ok I get it now, then so a new request is needed, should I wait for GP to reply (which could be awhile) or just do it?

PR #202, opened 2018-01-14
Cicada: Streamers Restoration, Addition, & Update

HellDragon-HK:
This restores the missing streamers as well as adds in the 2 additional ones the UT3 Cicada has.  Placement I think is fairly close but with the animation issues still present it's not going to be quite accurate later on but should do for now.

### DIFF
--- a/Classes/UT3Cicada.uc
+++ b/Classes/UT3Cicada.uc
@@ -90,6 +90,17 @@ simulated function DrivingStatusChanged()
                     TrailEffects[i].SetRelativeRotation( TrailEffectRotation );
                 }
         }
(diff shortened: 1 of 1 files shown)


GreatEmerald:
This needs to be fixed rather by removing code, not adding it, quite like in the other class. Replace `super(ONSChopperCraft).DrivingStatusChanged();` with `Super.DrivingStatusChanged();` and remove everything until `if (Driver == None)`, and this will work just as well.

HellDragon-HK:
That didn't work, it removed all streamers and caused engine flames to be vertical.

PR #203, opened 2018-01-14
Paladin: Smoke & Fire Locations

HellDragon-HK:
You know the drill, 2 places and 2 different sizes.  I think the front is probably the best option for a prelude.

### DIFF
--- a/Classes/UT3Paladin.uc
+++ b/Classes/UT3Paladin.uc
@@ -283,6 +283,11 @@ defaultproperties
     //Aerial View
     //TPCamWorldOffset = (X=0,Y=0,Z=200);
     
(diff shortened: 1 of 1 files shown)


GreatEmerald:
Hm, this pull showed a pretty large problem in one of your previous pulls that I didn't notice in time... A `Sound'` property was missing the closing `'`. Fixed now, but be careful about that...
Otherwise, OK.

HellDragon-HK:
I caught that too somewhere a few days ago, I believe it was the Cicada, it complied fine but didn't work in-game resulting in stock sound so not really a large problem causing complete instability as just a minor mistake.

PR #207, opened 2018-01-15
Hellbender: Suspension Changes

HellDragon-HK:
The corrected scale caused suspensions to not work as well so here are some values that seem better.  I've not had any luck finding a value for the SuspensionTravel for turn bouncing yet, I've tried really high values and it seems like they get ignored, I also tried WheelSoftness to the point the Hellbender spawned like gravity was really crushing it where it was into the ground and couldn't even be entered.


HellDragon-HK:
Forgot some important entries

GreatEmerald:
OK.

PR #210, opened 2018-01-16
Goliath: Various Updates

HellDragon-HK:


GreatEmerald:
You have broken the Goliath file by overwriting it (see https://travis-ci.org/GreatEmerald/UT3Vehicles). You need to re-overwrite it with the old file now so that the build succeeds, then upload your new file with a new name, and update `Binaries.csv` in this pull request.

HellDragon-HK:
Ver 1 is still there, uploaded Dec 19 2017, is that not the right one?

GreatEmerald:
The file https://drive.google.com/uc?export=download&id=1uFWoDVvJpaGgDzxAYA8KkGT6Ija4CRKI must match the checksum `f2c9cafc34d60ec4321d1402cee7e8345a31265f86a3e85667e6e31d77bd8ed0`. It is currently `7282b51bc6b321c39ee88b8ad8082d9313e8d5c2205498e7393c5058401c3a26`. Recheck your versions to see which one corresponds to which.

HellDragon-HK:
Ver 1 is f2c9cafc34d60ec4321d1402cee7e8345a31265f86a3e85667e6e31d77bd8ed0

GreatEmerald:
Then why is the URL pointing to some other (version of the) file?

HellDragon-HK:
The versions stack with the new one having the same url & name while the old one is still there in manage versions but I can't do much with it other than download and delete

GreatEmerald:
Yea, so that's the problem. You need to make sure the names are all unique, so that Google assigns new URLs to them. You can fix this by re-uploading Ver 1 with the same name as you used originally, which should make the URL point back to Ver 1. Then for Ver 2 upload it with a different name or in a different folder.

HellDragon-HK:
Which would be more preferable as a different name will also cause build issues won't it?

GreatEmerald:
No, names on your Google Drive don't matter at all. The `FetchBinaries.sh` script renames all files to the names in `Binaries.csv`, so that they are always correct.

HellDragon-HK:
Ok didn't know that, so I'll mark them with Ver# then from now on

GreatEmerald:
Yeap, that will work. Just make sure to fix the one without a version so it's Ver 1.

HellDragon-HK:
Rename doesn't break file paths so shouldn't be an issue

GreatEmerald:
I guess? Let me know when that URL points back to Ver 1 so I can retrigger the `Binaries.csv` consistency check.

HellDragon-HK:
I think it does now

GreatEmerald:
Yeap, it works now! Now the Hellbender sound file is mismatched. Let me know when you fix all the files.

Well, you did from the URL perspective, but not from the file perspective. Which makes it all good as long as you can fix the URLs again.

HellDragon-HK:
I think I can, just point them out as they come I guess, btw welcome back.

HellDragon-HK:
Hellbender should be fixed

GreatEmerald:
And there you go, Travis has picked up this Pull Request and is telling you that `Binaries.csv` is still inconsistent. If you press on Details and scroll to the very end, you can see which file failed.

I'll also set up a cache so that it doesn't redownload all files every time it tries to check it.

HellDragon-HK:
I really don't know how to read that mess GE, I can't see what it's complaining about.

GreatEmerald:
Yeap, the check passes now! All files are fixed, thanks.

HellDragon-HK:
I really really hate this Travis CI thing GE, everything I do it complains about which usually results in having to re-upload a file to get a new link and hope it passes, this time I can't even tell what it's complaining about.

PR #211, opened 2018-01-16
Update Binaries.csv

HellDragon-HK:


HellDragon-HK:
Attempted to improve Cicada volume again, now the only sounds having issues are explosion and missile fire, everything else seems like it might be good.   Missile fire is at 104 which is a lot higher than most other sounds that are already fine so I'm thinking it's something that needs to be tweaked in the code like radius maybe.   @GreatEmerald what do you think?

HellDragon-HK:
Still working on trying to perfect the Cicada uax and fixed some distorted sounds as pointed out by Alex.

HellDragon-HK:
Updated again to try and fix distortion issues @Unre-Alex  pointed out

Unre-Alex:
The Nightshade uax is incompatible with [#186](https://github.com/GreatEmerald/UT3Vehicles/pull/186)

HellDragon-HK:
Does that fix it?

Unre-Alex:
Yes

GreatEmerald:
You are breaking several rules here that you really need to keep in mind.

1. Never replace a file in your Google Drive if it has been put up in `Binaries.csv` and accepted into the `master` branch. Ever. Git relies on having all history of files so that we can revert to previous files if they are found to be broken. If you replace a file, it's gone. So if you make a new version of a file, upload it *alongside* the old version (you can give it a different name on your Google Drive and keep the usual name in `Binaries.csv`).
2. All edits to `Binaries.csv` must be made in pull requests where you are modifying paths to the new binaries (unless the binary update doesn't include any changes to paths). Otherwise your commits go out of sync and the package will fail to build.

In order to enforce the first point, I have now set up Travis CI solely to test whether `Binaries.csv` is valid. It must be valid at all times, retroactively. And for the second point, I'm closing this request, you need to submit changes to individual lines to your other pull requests instead.

HellDragon-HK:
Which file got deleted, because they do upload as new versions and I thought I was only removing ones that were not already accepted.

Do you mean add the binary change request to the request that needs them?

PR #212, opened 2018-01-16
Cicada: Various Updates

HellDragon-HK:
Here is the correct engine sound we hear in UT3....I have mixed feelings as I do like the non-modulated one a lot....shame we can't use both randomly which I tried a sound group and the result was complete silence.


HellDragon-HK:
Sound update to give correct sounds and I think I finally got the turret alt fire sound working too (non recorded now)
UT3 Bullet impact sounds added
Turret fire location corrected

PR #213, opened 2018-01-16
Hellbender: Various Updates

HellDragon-HK:
Well this engine sound is right but we've still got a bit of a pitch issue that I don't know how to solve, it does seem closer by adding in IdleRPM=2000 but not quite there yet, IdleRPM=4000 is about what we should hear for when actually moving I believe.


HellDragon-HK:
Sounds redone for UT3 style variance (non recorded now)
UT3 Bullet Impact sounds added
Didn't have `Drawscale = 1.0` in my class so a few things needed to be redone like the headlights

HellDragon-HK:
Fixed combo self damage problem

Unre-Alex:
Thanks for fixing the combos! Like with the suspension, a small fix for a big issue.

This is duplicated in the class:

    EntryPosition=(X=0,Y=0,Z=0)
    EntryRadius=180.0  //300.000000

HellDragon-HK:
Thanks for pointing that out Alex, and in light of that and you said the ImpactDamageMult went missing, I've done what I've been wanting to do to everything and try to organize it more....would like to make it uniform format across all the other classes too if possible so we no longer need to look all over the place to avoid this from happening, any objections or opinions?

If unclear what I mean by this is all classes have the same order of info in defaults so for the main body of the vehicle it would be the looks/visuals related first, then sounds, damage, movement, entry & exit, and then camera, this way we have less issues with duplicate entries because we don't need to look at everything in the defaults area to try and make sure there are none and will only need to check a specific section.

I had a lot of duplicate issues with the Viper recently because it's defaults are all over the place for example.

HellDragon-HK:
Ok done with reorganizing...or at least attempting it, if it's a good idea and it came out well I'll slowly do it across other classes as well, if it is a good idea but needs some more thought we can do that here too before doing others.

EDIT - Oh and in a completely unrelated note but because I need to mention it somewhere, the CollisionHeight & CollisionRadius entries have nothing to do with physical impact that I can see, they appear to be the requirements for controlling if spawning has enough room which I've been trying to use on the Cicada but no luck fixing it on Urban with it.

PR #214, opened 2018-01-16
Leviathan: Sound Update

HellDragon-HK:


HellDragon-HK:
TurretFire has a sound that I don't believe should be there that I think is an in-flight stock UT2004 sound the rockets have.

Tried the the Stinger Alt Fire sound on the Stinger Turret but it doesn't quite work right and only plays every few seconds instead of continuously like it should, this change is not included here.

HellDragon-HK:
Fixed Leviathan driver primary not homing

HellDragon-HK:
....And fixed shots not hurting non-vehicle targets......I don't understand how vehicles were taking full damage and pedestrians were taking 0

PR #215, opened 2018-01-16
Manta: Various Updates

HellDragon-HK:


HellDragon-HK:
This is everything I've got working for the Manta right now which includes...

-Updated sounds to include UT3 style variations which is set up for Basic or Extended uax files (don't think the Manta primary fire will sound right with Basic but you said only 3 sounds so there you go)
-UT3 Bullet Impact sounds which appear to work for hits from any weapons or vehicles
-Plasma Impact sound
-Engine loop closer to UT3
-Engine flames, fixed the jump problem by somewhat increasing JumpForceMag
-Guns movement (up & down only, can't get them to move side to side as well)
-UT3 visually damaged locations (where the smoke is coming from)
Think that's all

PR #216, opened 2018-01-16
Manta Plasma: Impact Sound

HellDragon-HK:
This sets up Manta plasma to have it's impact sound but doesn't seem to work, this code block is directly from the HoverBikePlasmaProjectile class which used BioRifle sound which I don't think I recall ever hearing in UT2004 so possibly this bit of code never worked right in default UT2004.


HellDragon-HK:
No longer WIP, I got it to work in-game

Unre-Alex:
It works! nice!

PR #218, opened 2018-01-20
64-bit Compatible Textures

Unre-Alex:


HellDragon-HK:
Cool Alex, what all has changed since the one I have?  I see you fixed up the Fury & Stealthbender, haven't tried it yet but will when I can.

Unre-Alex:
Nothing much that's noticeable with the Axon, just improved normal maps a bit and reduced a specular for the SPMA and Leviathan's wheels.

Unre-Alex:
added hoverboard

GreatEmerald:
For the Stealthbender, why do you have two `BlueSkinB`s?

Also, it would be nice if you could resubmit this pull request so that the Travis CI I just set up could validate your changes to `Binaries.csv`. You can close this request and then open a new one from the same branch (`Unre-Alex:patch-3`).

GreatEmerald:
Oh, it will also recheck it if you change anything in the code (such as those `BlueSkinB`/`BlueSkinA` variables).

Unre-Alex:
Two SkinBlueB? One is needed to show its 2-part skin when spawned, and the second when it's in use and uncloaked. If that's what you meant.

GreatEmerald:
I meent that you only ever use `[0]` for the B variants. Does it really need to be an array?

Unre-Alex:
That's how the Leviathan's skin is coded.

PR #220, opened 2018-01-24
Fury: Sound Update

HellDragon-HK:
The Fury has it's own sounds in UT3, in UT2004 currently it's just stock Raptor sounds so this fixes that.
There is no Fury UAX in the current archive so this requires a completely new binary file.

In case you see this before the other message GE, the Fury doesn't appear to have ANY explode sound at all even in-game UT3, so either we need to leave it as none or borrow from another vehicle. (Further proof UT3 wasn't finished isn't it?)


HellDragon-HK:
For an explosion sound I think decent choices are the Cicada, Nemesis, & Scorpion, I'm leaning toward the Cicada at the moment but what does everyone else think?

PR #221, opened 2018-01-25
Hoverboard: Sound Update

HellDragon-HK:
Engine & Jump closer to being correct as well as randomized sounds, still not as varied as UT3 however.


GreatEmerald:
The `Binaries.csv` update fails because your URL is in the wrong format.
(Travis was a bit bugged due to the cache, now fixed.)

PR #222, opened 2018-01-27
Nightshade: Sound Update

HellDragon-HK:
Volume may still be an issue especially with the engine loop


HellDragon-HK:
Nightshade doesn't appear to have an explosion sound, I believe the Viper's explosion sound would work decent though.

PR #223, opened 2018-01-28
Raptor: Sound Update

HellDragon-HK:
All seems decent here except perhaps engine loop but that's always the troublesome one it seems.


HellDragon-HK:
Updated sounds to for variations (non recorded now)
Added plasma impact sound
Added correct explosion sound for rocket
Added UT3 bullet impact sounds

HellDragon-HK:
@GreatEmerald Alex recommended disabling the Rocket's ambient sound because it extremely loud the whole time it's in flight like the sound is coming from the Raptor & not the rocket, but while I've confirmed the sound is coming from the rocket I'm still not so sure it isn't also coming from the Raptor too.

SoundRadius & TransientSoundRadius do nothing for the problem and only seem to be affecting the explosion sound.

HellDragon-HK:
This might be what's going on....

The sound is coming from the smoke trail, and in this case it's the ONSAvrilSmokeTrail so perhaps it's playing 2 sounds on top of one another while using the volume settings in the smoke trail class?

HellDragon-HK:
I give up on it, I've tried volume, I've tried a creating a new trail class with the right ambient, I've tried removing the ambient sound from one class and adding it to the other...what else is there to try?

I want to point out it can't be the sound itself because the Rocket Launcher doesn't have this problem, so far it's unique to the Raptor.

EDIT - The stock Raptor has this issue to so it's a bug on Epic's part apparently.

EDIT2 - Here is something weird, you only hear this if you're in the Raptor, jumping out it stops but if you get back in and the missile is still in flight you hear it again no matter how far away it is.

PR #224, opened 2018-01-30
Scorpion: Various Updates

HellDragon-HK:
All seems to work except can't find a way to test collide since the Scorpion isn't taking impact damage.


HellDragon-HK:
Updated to allow varied sounds
Added UT3 Bullet Impact sounds
Added Eject sound (borrowed from Viper code)
Fixed blades not breaking easily problem
Suspension tweaks to help give the bounce on turning effect

PR #233, opened 2018-02-05
64-bit Compatible Textures

Unre-Alex:


GreatEmerald:
What do you mean by that? If you remove all `[2]`, `[0]` and `(0)`s fromt he code, it should still work fine.

HellDragon-HK:
Wouldn't that return it to not having a texture for the necris tech part, which in beta 4 it uses the Hellbender body texture there and looks fairly bad in my opinion.

Unre-Alex:
Actually it works if they're removed, let me edit this then.

GreatEmerald:
OK, thanks!

PR #237, opened 2018-03-04
Textures for Deployables

Unre-Alex:
New textures and added missing ones (LinkStation, ShapeCharge, XRAY and team skins for the SpiderMine).

This also puts their textures, sounds and animations into separate files.


HellDragon-HK:
Everything seems to work but I'd recommend adding 2.0 to the PlaySound, like this PlaySound(DropSnd, SLOT_None, 2.0);

Without it on ONS-IslandHop which has no music and little to no ambient sound I couldn't hear the mine sounds until I added the 2.0, I also tried 1.5 but it was too quiet in my opinion so something between 2.0 & 1.5 I think.

I also think GreatEmerald is going to call for an indentation fix on the PlaySound you added, it looks a little off with the rest of the block.

I'm also not familiar with how the mines look in UT3 so I'm going by everything works here as it has a visible texture

Unre-Alex:
Even with 1.5 the sounds are too loud comparing to other sounds. It's good how it is now.

HellDragon-HK:
Something went wrong somewhere, not sure if it was here or somewhere else but while Instant Action runs fine, LAN will no longer load and no it's not the extended sound files as UT3Weapons loads....only thing I can see is it might have something to do with the Nightshade as I have a Missing Package UT3NightshadeTex which doesn't make sense as I can see it clearly in the textures folder :S

Running a dedicated server works but I get a black screen on trying to join it.

EDIT - Disregard, I got it working all though not sure what I did different but oh well at least no problems apparently.

PR #239, opened 2018-03-17
Missing Vehicles: Sounds for Later

HellDragon-HK:
Alex & I thought it would be good to get these done too that way if/when these missing vehicles are added the sounds will mostly be done and ready if not completely.

I haven't gotten the Scavenger done yet, I think that's all that is missing besides the utility Generic sounds files which also needs to be done.

Technically @100GPing100 should be able to use the Nemesis uax at least so it's not quite missing is it, more like somewhere in between.


HellDragon-HK:
Finally got the Scavenger done, I swear that one has more sounds than anything else in all of UT3

